PR #5, opened 2020-06-23
adds firebase messaging for app to receive messages; creates social worker js file to handle notifications when not active

nlumiere:


arringtonh (approved):
lgtm!

PR #6, opened 2020-06-23
write real time object data to HTML from firestore realtime database

arringtonh:
This PR allows changes to the firebase realtime database to be updated in the HTML in real time. Because of the way the objects are stored in the database, it is possible that we can use this feature to have real time chatting by using the timestamp in milliseconds as a key.


nlumiere (approved):
lgtm

PR #7, opened 2020-06-24
display messages from the datastore as a list

arringtonh:
This PR takes input from the user and displays it to them in real time (without refreshing the page) as a list


public/script.js
```diff
@@ -1,15 +1,34 @@
...
+    });
+}
+
+function pushChatMessage() {
+    const messageRef = firebase.database().ref(path);
+
+    const message = document.getElementById('message-input').value;
+    // push message to datastore
+    messageRef.push(message);
+    console.log("yo")
```
showermat: Is it possible to avoid merging debug code into master?  Ideally, by the time you make a pull request, the code should be debugged sufficiently that you don't need this any more.

showermat (approved):
One minor comment, but looks good.

public/script.js
arringtonh: you're right, i need to get better about that but I keep forgetting to take it out 🤕

PR #12, opened 2020-06-26
load more messages when the user scrolls to the top of the chat box

arringtonh:
the last commit was supposed to say "remove the debugging print statement from the top of the DOM' but I accidentally used the same commit statement twice


proj/src/main/webapp/index.html
```diff
@@ -27,6 +27,7 @@
 
         <script src="https://www.gstatic.com/firebasejs/7.15.4/firebase-database.js"></script>
         <script src="https://www.gstatic.com/firebasejs/7.15.4/firebase-messaging.js"></script>
+        <script src="app.js"></script> <!-- i put the initialization stuff in here so i don't have to take it out the commits every time -->
```
aokeke: I think the comment can be done in the PR rather than in the file. Or maybe in the file just have something like `<!--Holds initialization scripts -->`

proj/src/main/webapp/script.js
```diff
@@ -38,28 +35,34 @@ function appendMessage(payload){
...
+
 const path = '/messages'; // can make this more detailed (for example add user ID)
+const LIMIT = 20; // how many messages to load at a time
+var FIRST_CHILD_KEY;
 
 function init() {
     initRef();
+
+    const chat = document.getElementById('chat-as-list');
+    chat.addEventListener('scroll', addMoreMessagesAtTheTop);
```
aokeke, line 97: Interesting. I didn't know there was a scroll event listener. Does this happen every time you scroll or only when you scroll to the top? What if they are trying to scroll down even though they are the bottom?

proj/src/main/webapp/script.js
```diff
@@ -69,4 +72,35 @@ function pushChatMessage() {
...
+    }
+}
+
+function addMessagesToListElement(messages, firstChild, oldScrollHeight) {
+    const chat = document.getElementById('chat-as-list');
+    for (var key in messages) {
+        if (messages.hasOwnProperty(key)) {
+            if (!FIRST_CHILD_KEY) {
+                FIRST_CHILD_KEY = key;
+                continue;
```
aokeke: This can be an if-else rather than using continue.

proj/src/main/webapp/script.js
```diff
@@ -69,4 +72,35 @@ function pushChatMessage() {
     const message = document.getElementById('message-input').value;
     // push message to datastore
     messageRef.push(message);
-} 
+}
+
+var messages;
```
aokeke: Why is this. defined outside of all other functions? I don't see where we add to this or set this.

proj/src/main/webapp/script.js
```diff
@@ -38,28 +35,34 @@ function appendMessage(payload){
 /*
     Realtime Database
  */
+
 const path = '/messages'; // can make this more detailed (for example add user ID)
+const LIMIT = 20; // how many messages to load at a time
+var FIRST_CHILD_KEY;
```
aokeke: I think it is good that LIMIT is in all caps because it is a constant, but I think FIRST_CHILD_KEY isn't a constant since it gets set and mutated by other places, so maybe that one shouldn't be all caps.

aokeke (changes requested):

proj/src/main/webapp/script.js
arringtonh: it only happens when you scroll to the top, i use an if statement to check if they're at the top

proj/src/main/webapp/script.js
arringtonh: this was a debug statement i forgot to remove

proj/src/main/webapp/script.js

proj/src/main/webapp/script.js

proj/src/main/webapp/index.html
arringtonh: fixed!

proj/src/main/webapp/script.js
```diff
@@ -69,4 +72,33 @@ function pushChatMessage() {
     const message = document.getElementById('message-input').value;
     // push message to datastore
     messageRef.push(message);
-} 
+}
+
+function addMoreMessagesAtTheTop() {
+    const dbRefObject = firebase.database().ref(path);
+    const chat = document.getElementById('chat-as-list');
+    if (chat.scrollTop === 0) {
```
showermat, line 133: Have you considered loading when they get close to the top but before they actually reach it?  I don't know if that's the right thing to do, but it's worth considering.

nlumiere (approved):
cool

showermat (approved):

proj/src/main/webapp/script.js
arringtonh: I tried do that, but it ended up triggering the function more than once even when the tolerance was really small, so I just made it so it has to be 0

showermat: There are clever ways to deal with that, but I agree that's not a priority.

leeannachen:
lgtm

aokeke (approved):
lgtm

aokeke:
Please resolve merge conflicts before merging.

PR #14, opened 2020-06-26
Adds ugly authentication at top of page; does not impact functionality yet; confirmed users displayed on firebase console

nlumiere:


proj/src/main/webapp/index.html
```diff
@@ -11,13 +11,26 @@
         <!-- TODO: Add SDKs for Firebase products that you want to use
             https://firebase.google.com/docs/web/setup#available-libraries -->
         <script src="https://www.gstatic.com/firebasejs/7.15.4/firebase-analytics.js"></script>
+
+        <script src="app.js"></script>
```
showermat, line 15: Are you including a file that's gitignored?  Why?

proj/src/main/webapp/index.html
```diff
@@ -11,13 +11,26 @@
...
+        <div class="login-container">
+            <input id="email" type="email" placeholder="Email">
+            <input id="pass" type="password" placeholder="Password">
+            
+            <button id="btnLogin">Log in</button>
+            <button id="btnSignUp">Sign Up</button>
+            <button id="btnLogout" class="hidden">Log out</button>
+        </div>
+
+        <script src="script.js"></script>
```
showermat: Why isn't this script with all the other scripts above?

showermat (commented):
A couple questions before you merge.

proj/src/main/webapp/index.html
nlumiere: This is the file that has all of the private information required to run. We each have a version of the file, but aren't enthusiastic about having it publicly available

proj/src/main/webapp/index.html
arringtonh: this is also my question

nlumiere: This script needs to be after the authentication objects in order for it to run properly. less than Ideal, but this works until we discover a workaround.

showermat: Is it so that the const variables at the top can be initialized properly?  Can that be fixed by creating them in an init function and calling that when the page is done loading?  I think that would be better in general...although I don't really know what the best practices for JavaScript are.

proj/src/main/webapp/index.html
showermat: Ah.  But then when it's deployed, the user can just fetch the script from the server and see it that way, no?  Doesn't sound like a very good security model.  I think in general it's impossible to keep information in client-side JavaScript private, which is why it's important to have server-side checks that prevent users from abusing that information.  Do we have a plan to move forward so that this information can eventually be committed without a security concern?  If not, I think we need to make one.

proj/src/main/webapp/index.html
leeannachen: Script tags are normally in the head or bottom of the body (so the script runs after the page is pared and loaded). There are also the async and defer options to choose when to load the script.

proj/src/main/webapp/index.html
```diff
@@ -11,19 +11,32 @@
         <!-- TODO: Add SDKs for Firebase products that you want to use
             https://firebase.google.com/docs/web/setup#available-libraries -->
         <script src="https://www.gstatic.com/firebasejs/7.15.4/firebase-analytics.js"></script>
+
+        <script src="app.js"></script>
```
showermat, line 15: Is the problem of when to load this script fixed?

proj/src/main/webapp/style.css
```diff
@@ -59,4 +62,5 @@ h4 {
   line-height: 34px;
 
   color: #FFFFFF;
+>>>>>>> 967a1154d9c92417367fac4c3d0340d5c48f7174
```
showermat: It looks like a stray merge line is left in here.

showermat (changes requested):

proj/src/main/webapp/index.html
nlumiere: Yes, based on lee's comment, the initialization scripts are in the header while the logic scripts are at the bottom of body

showermat (approved):

PR #18, opened 2020-06-29
adds authentication functionality to sign in/sign up

nlumiere:


leeannachen (approved):
lgtm *insert thumbs up emoji

proj/src/main/webapp/script.js
```diff
@@ -3,53 +3,69 @@
...
+        firebase.auth().signOut();
+        window.location.replace("welcome.html");
+    });
+}
 
 firebase.auth().onAuthStateChanged(firebaseUser => {
     if(firebaseUser){
         console.log(firebaseUser);
-        btnLogout.classList.remove("hidden");
+        if(btnLogout)
```
arringtonh, line 62: out of curiosity what does adding this if statement do? is it checking if it's hidden or not?

proj/src/main/webapp/script.js
```diff
@@ -3,48 +3,69 @@
...
+    btnSignUp.addEventListener("click", e => {
+        const emailVal = txtEmail.value;
+        const passVal = pass.value;
+
+        // Initialize auth object
+        const auth = firebase.auth();
+        auth.useDeviceLanguage();
+
+        const promise = auth.createUserWithEmailAndPassword(emailVal, passVal).then(function(){
+            auth.currentUser.updateProfile({displayName: fname.value + " " + lname.value}).then(function(){
```
arringtonh, line 40: here i think you forgot to initialize lname because i don't see it anywhere else in this file

arringtonh (commented):
welcome.html and signup.html look nearly identical, is that what lee was talking about earlier with having separate pages for mobile?

proj/src/main/webapp/script.js
nlumiere: The old code removes a class "hidden" to the element, which switches its visibility state
The new code checks to see if it exists

PR #19, opened 2020-06-29
Update code to reflect new datastore storage heirarchy.

arringtonh:
The structure of the datastore now goes chat -> other metadata (chat name, chat tag) + chat messages -> content of message + other metadata (time stamp, user ID once we get that working). This will make it easier for us to get information about the chat and the messages, make multiple distinct chat rooms, and add other metadata as we see fit.

### DIFF
--- a/proj/src/main/webapp/script.js
+++ b/proj/src/main/webapp/script.js
@@ -85,7 +85,8 @@ function appendMessage(payload){
     Realtime Database
  */
 
-const PATH = '/messages'; // can make this more detailed (for example add user ID)
+const CHAT_ID = '-MB0ycAOM8VGIXlev5u8'
+const PATH = '/chat/'+CHAT_ID+'/messages'; // can make this more detailed (for example add user ID)
 const LIMIT = 20; // how many messages to load at a time
 var firstChildKey;
 
@@ -110,18 +111,23 @@ function initRef() {
             firstChildKey = snap.key;
         } else {
             const li = document.createElement('li');
-            li.innerText = snap.val();
+            li.innerText = snap.val().content;
             listObject.appendChild(li);
         }
     });
 }
 
 function pushChatMessage() {
-    const messageRef = firebase.database().ref(PATH);
-
     const messageInput = document.getElementById('message-input');
+
+    const chatRef = firebase.database().ref(PATH);
+
+    var message = {
+        content : messageInput.value,
+        timestamp : new Date().getTime()
+    }
     // push message to datastore
-    messageRef.push(messageInput.value);
+    chatRef.push(message);
     messageInput.value = null; // clear the message
 }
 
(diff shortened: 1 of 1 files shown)


proj/src/main/webapp/script.js
```diff
@@ -110,18 +111,23 @@ function initRef() {
             firstChildKey = snap.key;
         } else {
             const li = document.createElement('li');
-            li.innerText = snap.val();
+            li.innerText = snap.val().content;
```
nlumiere, line 114: really happy about this

proj/src/main/webapp/script.js
```diff
@@ -85,7 +85,8 @@ function appendMessage(payload){
     Realtime Database
  */
 
-const PATH = '/messages'; // can make this more detailed (for example add user ID)
+const CHAT_ID = '-MB0ycAOM8VGIXlev5u8'
+const PATH = '/chat/'+CHAT_ID+'/messages'; // can make this more detailed (for example add user ID)
```
nlumiere, line 89: is this going to remain hardcoded? Also are all chats going to be under the /chat directory?

proj/src/main/webapp/script.js
```diff
@@ -110,18 +111,23 @@ function initRef() {
...
-    const messageRef = firebase.database().ref(PATH);
-
     const messageInput = document.getElementById('message-input');
+
+    const chatRef = firebase.database().ref(PATH);
+
+    var message = {
+        content : messageInput.value,
+        timestamp : new Date().getTime()
+    }
```
nlumiere, line 128: Cool, thanks for starting this. Once I have auth more up and running I'll add relevant user data to message

proj/src/main/webapp/script.js
arringtonh: the chat ID won't remain hardcoded but the rest of the path will. And yeah all the chats will be under /chat

proj/src/main/webapp/script.js
leeannachen: ^^ i am assuming we can also put profile img for each user in the future

leeannachen (commented):
lgtm 👍

proj/src/main/webapp/script.js
aokeke: I think we can make `PATH` into a function like `getPath` that takes `CHAT_ID` as an input. 

Also I'm not sure if you want to store userId and stuff in the path because then it means you'd have to store all the messages once for each user that has access to them, unless there is a way to have multiple paths pointing at the same data with firebase.

proj/src/main/webapp/script.js
aokeke: I'm not sure we'd want to store profile image here because then we'd be having an image stored with each message. I think maybe storing username and then calling the server to get the profile images for the username might help reduce the number of places you need to store the same data.

aokeke (approved):
LGTM

proj/src/main/webapp/script.js
arringtonh: yeah userID won't be stored in the path, it will just be stored as metadata for each message

proj/src/main/webapp/script.js
arringtonh: ^ what arinze said, i think we should store profile picture and display name with the user data instead of the message itself

PR #20, opened 2020-06-30
Revert "adds authentication functionality to sign in/sign up" because I didn't realize matt and arinze did not review

nlumiere:
Reverts googleinterns/step14-2020#18


proj/src/main/webapp/script.js
```diff
@@ -3,69 +3,48 @@
...
-                window.location.replace("chat.html");
-            }).catch(function(){
-                console.log("error updating display name");
-            })
-        });
-        promise.catch(e => console.log(e.message));
+// Add sign up event 
+btnSignUp.addEventListener("click", e => {
+    const emailVal = txtEmail.value;
+    const passVal = pass.value;
```
aokeke, line 27: I'd expect there to be a confirm password field when signing up that you make sure is the same as the password field.

proj/src/main/webapp/static/signup.html
```diff
@@ -55,18 +41,17 @@ <h4>welcome</h4>
...
-              <input type="text" id="fname" name="fname" class="input"><br>
+              <input type="text" id="fname" name="fname" class="input" required><br>
               <label for="lname">last name:</label><br>
-              <input type="text" id="lname" name="lname" class="input"><br>
+              <input type="text" id="lname" name="lname" class="input" required><br>
               <label for="email">email:</label><br>
               <input type="text" id="email" name="email" class="input" required><br>
               <label for="password">password:</label><br>
-              <input type="password" id="pass" name="password" class="input" required><br>
+              <input type="text" id="password" name="password" class="input" required><br>
```
aokeke, line 50: Shouldn't this be `type="password"`? https://www.w3schools.com/tags/att_input_type_password.asp

proj/src/main/webapp/static/signup.html
```diff
@@ -55,18 +41,17 @@ <h4>welcome</h4>
           <div class="row justify-content-md-center">
             <form>
               <label for="fname">first name:</label><br>
-              <input type="text" id="fname" name="fname" class="input"><br>
+              <input type="text" id="fname" name="fname" class="input" required><br>
               <label for="lname">last name:</label><br>
-              <input type="text" id="lname" name="lname" class="input"><br>
+              <input type="text" id="lname" name="lname" class="input" required><br>
```
aokeke, line 46: It looks like First Name and Last Name are not being used when the sign up button is clicked. Should they be?

aokeke (changes requested):

proj/src/main/webapp/static/signup.html
aokeke: Nevermind I see it in the other PR.

proj/src/main/webapp/static/signup.html
leeannachen: It should be. I updated it in a recent pull request.

PR #21, opened 2020-06-30
gives functionality to login/signup forms; saves custom user data to database on sign up; reroutes to chat.html while logged in as currentuser

nlumiere:


arringtonh (approved):
lgtm

proj/src/main/webapp/script.js
```diff
@@ -29,6 +30,7 @@ if(btnLogin){
 // Add sign up event
 if(btnSignUp){
     btnSignUp.addEventListener("click", e => {
+        const dbRef = firebase.database().ref("/users");
```
aokeke: I don't think this is being used.

proj/src/main/webapp/script.js
```diff
@@ -37,12 +39,23 @@ if(btnSignUp){
...
 
         const promise = auth.createUserWithEmailAndPassword(emailVal, passVal).then(function(){
-            auth.currentUser.updateProfile({displayName: fname.value + " " + lname.value}).then(function(){
+            const user = auth.currentUser;
+            user.updateProfile({
+                displayName: fname.value + " " + lname.value
+                }).then(function(){
                 console.log("display name updated successfully");
-                window.location.replace("chat.html");
+                if(fname && lname && tagStr){
```
aokeke: Does this mean we don't want to store tagStr if they didn't enter a last name for example? Or that we don't want to store their name if they didn't enter any tags? Maybe you can just set the fields that are present, or if all is required, i'd do some kind of validation before creating the user.

We previously set their display name using their fname and lname so I would think up there we might want to also check if fname and lname is populated.

aokeke (changes requested):

proj/src/main/webapp/script.js
```diff
@@ -37,12 +38,21 @@ if(btnSignUp){
...
-                window.location.replace("chat.html");
+                firebase.database().ref("users/" + auth.currentUser.uid).set({
+                    firstName : fname.value,
+                    lastName : lname.value,
+                    tags : tagStr.value.split(',')
+                }).then(function(){
+                    window.location.replace("chat.html");
+                });
             }).catch(function(){
                 console.log("error updating display name");
```
aokeke, line 54: Normally I'd suggest removing these console debug logs, but in this case, maybe you want to log something more descriptive while you debug. Like if the catch function can take a parameter with the error, you can log the error given.

In the finished product, you might want some sort of pop up with a user friendly error that says that an error occurred and suggesting they try again or something.

aokeke (approved):
LGTM 1 comment.

proj/src/main/webapp/script.js
nlumiere: Yup, that's what I was planning. Sounds good!

PR #23, opened 2020-06-30
Add more structure to comment format.

arringtonh:
Before, our comments were in a list which would make it difficult to add other parts to them (name, timestamp, etc) later. Now they are formatted with div elements which are much more flexible. I also update the css and HTML to reflect the new changes.


leeannachen (commented):
lgtm

proj/src/main/webapp/script.js
```diff
@@ -117,15 +117,15 @@ function appendMessage(payload){
...
+const PATH = '/chat/test/'+CHAT_ID+'/messages'; // can make this more detailed (for example add user ID)
 const LIMIT = 20; // how many messages to load at a time
 var firstChildKey;
 
 function init() {
     initRef();
     clickWithEnterKey();
 
-    const chat = document.getElementById('chat-as-list');
+    const chat = document.querySelector('.chat');
```
aokeke: I think if there is only one item, maybe getting it by Id will be better. If there are multiple chat divs on the page will you need to add this to each one? If so, you might need a loop below.

proj/src/main/webapp/script.js
```diff
@@ -117,15 +117,15 @@ function appendMessage(payload){
  */
 
 const CHAT_ID = '-MB0ycAOM8VGIXlev5u8'
-const PATH = '/chat/'+CHAT_ID+'/messages'; // can make this more detailed (for example add user ID)
+const PATH = '/chat/test/'+CHAT_ID+'/messages'; // can make this more detailed (for example add user ID)
```
aokeke: Is this just for testing. Should this be removed? If not, can you add a `TODO:` to remove it later?

aokeke (changes requested):

proj/src/main/webapp/script.js
arringtonh: so at some point, instead of test it would be a tag name, like running or books, /test/ is just what we're using for development. but yeah i added a todo

proj/src/main/webapp/script.js
arringtonh: since I don't think we know yet if we're going to display multiple chats at the same time, i just gave the one we have an ID and changed it to getElementByID instead

proj/src/main/webapp/script.js
```diff
@@ -205,3 +204,17 @@ function clickWithEnterKey() {
         }
     });
 }
+
+function createMessageWithTemplate(messageObj) {
+    const messageTemplate = document.getElementById('message-temp');
+    const message = messageTemplate.content.cloneNode(true);
+
+    const msgHeader = message.querySelector('.message-header');
+    const username = document.createElement('p');
```
showermat, line 213: Why are you creating this in the script rather than including it in the template?

showermat: Why are you creating this in the script rather than including it in the template?

showermat (changes requested):

proj/src/main/webapp/script.js
```diff
@@ -117,15 +117,16 @@ function appendMessage(payload){
  */
 
 const CHAT_ID = '-MB0ycAOM8VGIXlev5u8'
-const PATH = '/chat/'+CHAT_ID+'/messages'; // can make this more detailed (for example add user ID)
+// TODO: not hardcode, use path to an actual tag
+const PATH = '/chat/test/'+CHAT_ID+'/messages';
```
aokeke: Actually, can you make a getPath function that takes a chat id as a function so that in the future when chat id isn't static, it will be easier to change the path.

aokeke (changes requested):

proj/src/main/webapp/script.js

proj/src/main/webapp/script.js
arringtonh: No clue why I did that. fixed!

proj/src/main/webapp/script.js
```diff
@@ -134,16 +134,15 @@ function initRef() {
     // create database reference
     const dbRefObject = firebase.database().ref(PATH);
```
aokeke: can you call getPath here instead of using a constant.

proj/src/main/webapp/script.js
```diff
@@ -164,7 +163,7 @@ function pushChatMessage() {
 
 function addMoreMessagesAtTheTop() {
     const dbRefObject = firebase.database().ref(PATH);
```
aokeke: here too.

proj/src/main/webapp/script.js
```diff
@@ -117,15 +117,15 @@ function appendMessage(payload){
  */
 
 const CHAT_ID = '-MB0ycAOM8VGIXlev5u8'
-const PATH = '/chat/'+CHAT_ID+'/messages'; // can make this more detailed (for example add user ID)
+const PATH = getPath("test", CHAT_ID);
```
aokeke: Since you have a function to return path, I think you can remove this constant now.

proj/src/main/webapp/script.js
```diff
@@ -205,3 +203,19 @@ function clickWithEnterKey() {
...
+
+    const msgHeader = message.querySelector('.message-header');
+    msgHeader.getElementById('username').innerText = 'name go here';
+            
+    const msgBody = message.querySelector('.message-body');
+    msgBody.innerText = messageObj.content;
+    return message;
+}
+
+function getPath(tag, chatId) {
```
aokeke: to keep using the test tag, you can override tag to be 'test' in here so that you don't need to put `getPath('test',CHAT_ID)` everywhere and you can just create placeholder variables for tag where you are calling this.

proj/src/main/webapp/script.js
aokeke: Maybe something like `var tag; path=getPath(tag,CHAT_ID); const dbRefObject = firebase.database().ref(path);` That way later you just need to get tag and chat_id from some client side cached data based on what the user clicked on to get to the page.

aokeke (changes requested):

proj/src/main/webapp/script.js
arringtonh: what if instead of returning the path this function just returned the reference? that way it kills two birds with one stone

aokeke: That also works really well.

proj/src/main/webapp/script.js
```diff
@@ -116,55 +116,49 @@ function appendMessage(payload){
     Realtime Database
  */
 
-const CHAT_ID = '-MB0ycAOM8VGIXlev5u8'
-const PATH = '/chat/'+CHAT_ID+'/messages'; // can make this more detailed (for example add user ID)
+var CHAT_ID = '-MB0ycAOM8VGIXlev5u8'
+var tag = 'test';
+var dbRefObject = getDbRef(tag, CHAT_ID);
```
aokeke, line 121: Don't forget the `TODO:` to actually set the correct tag and move the getDbRef call to where it is actually needed.

aokeke (approved):

showermat (approved):

showermat:
Sorry for the delay!

PR #24, opened 2020-06-30
Makes multiple chat rooms to be populated with messages later

nlumiere:
upon user creation, adds user to chat rooms for each of their tags, then saves the user's data to the database along with their tags and a corresponding key for the associated chat.


arringtonh (commented):
lgtm!

proj/src/main/webapp/script.js
```diff
@@ -27,11 +27,64 @@ if(btnLogin){
     });
 }
 
+// Create or join chatroom
+function createOrJoinChat(currentTag){
```
aokeke, line 82: There is a lot being done here within anonymous functions. Can you break them out into their own function so it is a bit easier to follow what is going on.

proj/src/main/webapp/script.js
```diff
@@ -46,9 +99,13 @@ if(btnSignUp){
                 firebase.database().ref("users/" + auth.currentUser.uid).set({
                     firstName : fname.value,
                     lastName : lname.value,
-                    tags : tagStr.value.split(',')
+                    tags : tagList
```
aokeke: In addition to storing what users are in each chat, you can store chat keys or something in the user. Or a map from tags to the chat keys so that you don't need to go through all the users for each chat. This might be what arrington already said.

aokeke (changes requested):

proj/src/main/webapp/script.js
nlumiere: ya i'll get there soon

proj/src/main/webapp/script.js
showermat: Ah, I remember the days when I would put all my logic in a single function nested twelve layers deep.  Alas, those days are gone now....

showermat: Yeah, I'm not totally sure how this function is supposed to work.  Breaking it up might help.

showermat (changes requested):
Could you please break up `createOrJoinChat` and maybe document the pieces a bit so that I can tell what it's supposed to be doing?

proj/src/main/webapp/script.js
```diff
@@ -27,11 +27,72 @@ if(btnLogin){
...
+            query.once("value").then(function(snapshot){
+
+                snapshot.forEach(function(childSnapshot){
+
+                    var currentReference = childSnapshot.ref;
+                    if(childSnapshot.hasChild("users")){
+
+                        var userSnapshot = childSnapshot.child("users");
+                        var usersReference = userSnapshot.ref;
+                        if(userSnapshot.numChildren() < 200){
```
showermat: I'm not familiar with the Firebase libraries, so I'm not sure what this is, but it looks suspicious.  Is it limiting chat room size to 200?  If so, the 200 should be placed in a constant near the top of the file.  Does it fail silently when the condition is not met?  That sounds undesirable.

aokeke: I think they were planning to have the person join the room if it is less than 200 people and create a new subroom if it is greater, but yes I agree this should be in a constant near the top.

proj/src/main/webapp/script.js
```diff
@@ -27,11 +27,72 @@ if(btnLogin){
...
+                            addUserToTag(usersReference);
+                            return;
```
aokeke: I see, return if you've added the user, otherwise you will finish cycling through the list of chats and create a new chat below.

proj/src/main/webapp/script.js
```diff
@@ -27,11 +27,72 @@ if(btnLogin){
...
+        if(snapshot.hasChild(currentTag)){
+
+            // If tag already exists, navigate to users and add users if there is room
+            var query = firebase.database().ref("/chat/" + currentTag + "/").orderByKey();
+            query.once("value").then(function(snapshot){
+
+                snapshot.forEach(function(childSnapshot){
+
+                    var currentReference = childSnapshot.ref;
+                    if(childSnapshot.hasChild("users")){
```
aokeke: Will this ever not have a users node, and if so, is that an issue? Should we create a users node if one doesn't exist?

aokeke (changes requested):
This is looking a lot better. A couple lingering questions

proj/src/main/webapp/script.js
```diff
@@ -27,17 +29,101 @@ if(btnLogin){
...
+// Creates a new chat given a tag and adds the current user as a member
+function createNewChatWithUser(tag){
+    console.log("creating new chat with tag: " + tag);
+    var newChat = {
+        "name" : tag,
+        "tag" : tag,
+    };
+    var currentReference = firebase.database().ref("/chat/" + tag);
+    var postKey = currentReference.push(newChat).key;
+    currentReference = firebase.database().ref("/chat/" + tag + "/" + postKey + "/users/");
```
aokeke, line 49: Didn't arrington make a helper method for concatenating the currentTag and returning the reference or something close? I think it might be good to use that here so that if the path to chats changes (e.g. `/chat/tag/key/users/` -> `/messages/tag/key/users/` you will only need to change it in one place.

proj/src/main/webapp/script.js
```diff
@@ -27,17 +29,101 @@ if(btnLogin){
...
+
+// Create or join chatroom
+function createOrJoinChat(currentTag){
+    var ref = firebase.database().ref("/chat/");
+    return ref.once("value").then(function(snapshot){
+        // Checks to see if tag already exists in database
+        if(snapshot.hasChild(currentTag)){
+
+            // If tag already exists, navigate to users and add users if there is room
+            var query = firebase.database().ref("/chat/" + currentTag + "/").orderByKey();
```
aokeke, line 89: Maybe the same comment here depending on what the helper method was. Or maybe you can make a similar helper method for getting the reference to the tag and then the method for getting the reference to the key can build off of it.

proj/src/main/webapp/script.js
```diff
@@ -27,17 +29,101 @@ if(btnLogin){
...
+        console.log("key1 " + key);
+
+        var userSnapshot = childSnapshot.child("users");
+        var usersReference = userSnapshot.ref;
+
+        // Handled below if spillover is needed
+        if(userSnapshot.numChildren() < MAX_CHAT_SIZE){
+            addUserToTag(usersReference);
+            foundOpenRoom = true;
+            return true;
```
aokeke, line 70: Maybe it is a task for a subsequent PR so this one doesn't get even more complicated, but what if the user already exists in a chatroom? We might want to check that we are not adding them twice or adding them to a second room. Also I just learned that a reference is an acceptable datatype. https://firebase.google.com/docs/firestore/manage-data/data-types
I wonder if you can do something like have the list of users just be a list of references to the user objects or maybe the users tag map will map tag to a reference to the chat or something. But again, that is a bit more complex and probably better as a follow up PR.

proj/src/main/webapp/script.js
```diff
@@ -46,15 +132,15 @@ if(btnSignUp){
                 firebase.database().ref("users/" + auth.currentUser.uid).set({
                     firstName : fname.value,
                     lastName : lname.value,
-                    tags : tagStr.value.split(',')
+                    allTags : allTags
                 }).then(function(){
+
```
aokeke, line 137: Can you delete this extra line.

aokeke (changes requested):

showermat (approved):
Much easier to understand!

proj/src/main/webapp/script.js
nlumiere: Okay so I get what you're saying, but this seems pretty minor, and would not save terribly much. Either way, I'm providing a new path and would need to change it in all places that I'm getting a new reference; either as part of concatenation or as parameters to a function call.

showermat: Arinze, I know you like having a lot of helper methods in your code, but I think this is all right.

aokeke (approved):
lgtm

PR #28, opened 2020-07-01
Add sidebar that shows a user's most recently active chats, in order of last active; add ability to switch between user's chatrooms

arringtonh:
The plan is to add an onclick element to these divs that changes them to that chat.


arringtonh:
users will now to switched to whatever chat they click in the sidebar 🥳

aokeke (approved):
LGTM

leeannachen (approved):
cool! lgtm

PR #29, opened 2020-07-06
Fix issue where dupliacte messages showed up if you switched between …

arringtonh:
…chatrooms multiple times.


leeannachen (commented):
Small but mighty change 👍

aokeke (approved):
LGTM

PR #32, opened 2020-07-07
Add a sidebar with user information

arringtonh:


proj/src/main/webapp/script.js
```diff
@@ -344,12 +353,10 @@ function makeChatPreview(name, messageObj, tag, chatId) {
 }
 
 function populateSidebar() {
-    // Initialize auth object
-    const auth = firebase.auth();
-
+    // const auth = firebase.auth();
     // const iud = auth.currentUser.uid;
     const iud = "testuserwithdict";
```
aokeke: Should this still be the test user or the real user?

proj/src/main/webapp/script.js
```diff
@@ -366,4 +373,54 @@ function changeChatOnClick(domElement, tag, chatId) {
...
     });
+}
+
+function populateProfileSidebar() {
+    firebase.auth().onAuthStateChanged(function(user) {
+        if (user) {
+        const userRef = firebase.database().ref('/users/'+user.uid)
+        userRef.once('value', snap => {
+            const userObj = {};
+            userObj.photo = user.photoUrl;
```
aokeke, line 400: I'm not entirely sure how images are stored in firebase. Are they stored as a URL or as some sort of blob or something? I guess it depends on if you make users upload images or specify a URL to a photo already existing on the web to use

proj/src/main/webapp/script.js
```diff
@@ -366,4 +373,54 @@ function changeChatOnClick(domElement, tag, chatId) {
...
+    sidebar.classList.add("display-none");
+    settings.classList.remove("display-none");
+}
+
+function closeSettings() {
+    const settings = document.getElementById('user-profile');
+    const sidebar = document.getElementById('sidebar');
+
+    settings.classList.add("display-none");
+    sidebar.classList.remove("display-none");
```
aokeke: It looks like you're closing the settings and opening the sidebar? Should this sidebar always be open? Is the sidebar a list of all your chats and should it always be re-opened? If it is collapsible and it was collapsed before you clicked on settings, I would think it should stay collapsed.

proj/src/main/webapp/script.js
```diff
@@ -366,4 +373,54 @@ function changeChatOnClick(domElement, tag, chatId) {
...
+        }
+    }
+}
+
+function switchToSettings() {
+    const settings = document.getElementById('user-profile');
+    const sidebar = document.getElementById('sidebar');
+
+    sidebar.classList.add("display-none");
+    settings.classList.remove("display-none");
```
aokeke: Rather than using display-none as a psuedo css setting, can you do something like `settings.style.display="none"; settings.style.display="";`

aokeke (changes requested):

proj/src/main/webapp/script.js
arringtonh: it's still the test user, nick said he was making a PR to change that i believe

proj/src/main/webapp/script.js
arringtonh: yep, they're stored as a URL and saved to the firebase auth database

proj/src/main/webapp/script.js
arringtonh: from the mockups that Lee showed us I think we want the sidebar to always be open, but I can double check

proj/src/main/webapp/script.js

aokeke (approved):
LGTM

proj/src/main/webapp/script.js
showermat: Is this supposed to be "uid"?  "IUD" generally means something rather different.

nlumiere: I am indeed.

nlumiere (approved):
looks perf

proj/src/main/webapp/static/chat.html
```diff
@@ -27,7 +27,17 @@
     <body onload="init()">
         <div class="container screen chat-page full-height">
             <div class="row">
+
             <div class="col">
+
+                <aside id="user-profile" class="display-none">
```
showermat: Maybe I'm a little late saying this, but it feels weird to me to have a class just to set display: none.  Classes usually correspond to the kind of thing that needs to be styled (title, sidebar, chat message) rather than particular styling details (color, padding, display).  Is there a reason there's a class for this rather than just setting `style="display: none"`?

showermat (commented):
Looks reasonable; just a couple things I was confused about.

leeannachen (commented):
This is an edit function for the profile?

proj/src/main/webapp/script.js
arringtonh: yep... it seems i've made a very unfortunate spelling error

proj/src/main/webapp/static/chat.html
arringtonh: fixed!

PR #33, opened 2020-07-08
Personalizes onload chat content

nlumiere:
Sets previously-hard-coded displayed chat information such as chat content and headers to be specific to the logged in user

### DIFF
--- a/proj/src/main/webapp/script.js
+++ b/proj/src/main/webapp/script.js
@@ -234,18 +234,49 @@ var firstChildKey;
 
 function init() {
     const auth = firebase.auth();
-
-    clickWithEnterKey();
-    auth.onAuthStateChanged(firebaseUser => {
+    auth.onAuthStateChanged(async firebaseUser => {
         if(firebaseUser){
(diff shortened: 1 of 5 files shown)


showermat (approved):
Looks good to me!  Since `CHAT_ID` is no longer constant, you may want to lower-case it.

proj/src/main/webapp/script.js
```diff
@@ -23,7 +21,7 @@ if(btnLogin){
         const auth = firebase.auth();
 
         const promise = auth.signInWithEmailAndPassword(emailVal, passVal).then(function(user){
-            window.location.replace("static/chat.html");
+            window.location.replace("chat.html");
```
leeannachen, line 24: how come it doesn't need to go to the static folder to get the chat.html?

nlumiere: because both welcome.html and signup.html are already in the static folder

leeannachen: gotcha

PR #35, opened 2020-07-09
Allow user to change bio in sidebar on double-click

arringtonh:


leeannachen (commented):
sorry some names would need to be adjusted after merging; other than that, lgtm

proj/src/main/webapp/script.js
```diff
@@ -428,6 +428,35 @@ function addUsernameToMessage(uid, elementId) {
     })
 }
 
+const bioBox = document.getElementById('user-bio');
+const editInputBox = document.getElementById('bio-edit');
+
+bioBox.addEventListener('dblclick', function() {
+    const bioBox = document.getElementById('user-bio');
```
aokeke, line 441: You can also do 
`bioBox.addEventListener('dblclick', function(event) {
    const bioBox = event.target;}` But it looks like `this` might also give you a reference to `bioBox` so you don't need to search the document for it.

proj/src/main/webapp/script.js
```diff
@@ -428,6 +428,35 @@ function addUsernameToMessage(uid, elementId) {
     })
 }
 
+const bioBox = document.getElementById('user-bio');
+const editInputBox = document.getElementById('bio-edit');
```
aokeke: Is this and the functions below wrapped in some init function or are they just in the script. It's hard to tell with this diff tool. I think it would be better to put them in an onInit type function rather than defining constants in the middle of the script. Or defining all constants at the top is an alternative solution so they are all visible and don't get lost in the code.

aokeke (changes requested):

proj/src/main/webapp/script.js
arringtonh: fixed@

proj/src/main/webapp/script.js
arringtonh: fixed

proj/src/main/webapp/script.js
```diff
@@ -429,6 +437,36 @@ function addUsernameToMessage(uid, elementId) {
...
+function initBio() {
+    const bioBox = document.getElementById('user-bio');
+    const editInputBox = document.getElementById('bio-edit');
+
+    bioBox.addEventListener('dblclick', function() {
+        this.hidden = true;
+        const bio = this.innerText;
+
+        editInputBox.hidden = false;
+        editInputBox.innerText = bio;
```
aokeke: You could remove `bio` as the middle man and just do `editInputBox.innerText = this.innerText;` as you did on the line below this.

aokeke (approved):
1 tiny change, but I won't let it block the LGTM

proj/src/main/webapp/script.js
arringtonh: actually i just realized that that line of code doesn't do anything so i just removed it

PR #38, opened 2020-07-10
Make separate redirect function for index.html

arringtonh:


proj/src/main/webapp/index.html
```diff
@@ -18,7 +18,7 @@
         <script src="https://www.gstatic.com/firebasejs/7.15.4/firebase-database.js"></script>
         <script src="https://www.gstatic.com/firebasejs/7.15.4/firebase-messaging.js"></script>
     </head>
-    <body onload="init()">
+    <body onload="indexInit()">
```
aokeke: If all index is doing is redirecting with the tiny chunk you wrote, you can probably just put that script directly into the html and remove everything else from the body of this html file.

aokeke (changes requested):

proj/src/main/webapp/index.html
arringtonh: fixed!

aokeke (approved):

PR #41, opened 2020-07-13
Clear nonactive code 

leeannachen:


aokeke (approved):
LGTM. I don't know too much about file naming conventions, but I'm wondering if `chat.css` would be a better name than `chatstyles.css` That way the html and css have the same base name and it will be easier to find which CSS is affecting which html files.

proj/src/main/webapp/script.js
```diff
@@ -252,12 +252,6 @@ function init() {
     const chat = document.getElementById('chatbox');
     chat.addEventListener('scroll', addMoreMessagesAtTheTop);
 
-    const close = document.getElementById('close-settings');
-    close.addEventListener('click', closeSettings);
-
-    const settings = document.getElementById('settings-button');
-    settings.addEventListener('click', switchToSettings);
-
```
nlumiere, line 260: Is this now outdated? I thought this part was good. I didn't write it, I just don't know

showermat (approved):
Thanks!  Routine cleanup of the code base is important.  I agree with Arinze's comment about naming, though.

aokeke (changes requested):
Changing this to a requested change so that when other pieces move like the .js in arrington's PR, it will be consistent.

proj/src/main/webapp/script.js
leeannachen: There isn't a problem with the actual code; it's just that those functions are not needed anymore because the closing setting is replaced by collapsing the sidebar and the switching to setting is replaced by the settings drop down menu

leeannachen:
> LGTM. I don't know too much about file naming conventions, but I'm wondering if `chat.css` would be a better name than `chatstyles.css` That way the html and css have the same base name and it will be easier to find which CSS is affecting which html files.

gotcha; changing that right now

aokeke (approved):
LGTM

PR #43, opened 2020-07-13
Add a timestamp to messages and chat previews.

arringtonh:
Stretch goal: relative time stamps


aokeke (changes requested):
Similar to the comment in Lee's PR, can this file be called `chat.js`

It is a bit difficult to tell what changed vs what just moved. I think in the future can you have at least 2 separate commits. One for the moving of the content to the new file and one for adding the functionality.

proj/src/main/webapp/db.js
```diff
@@ -0,0 +1,302 @@
...
+    const bioBox = document.getElementById('user-bio');
+    const editInputBox = document.getElementById('bio-edit');
+
+    bioBox.addEventListener('dblclick', function() {
+        this.hidden = true;
+
+        editInputBox.hidden = false;
+        editInputBox.value = this.innerText
+
+        editInputBox.focus();
```
leeannachen: i am not sure if this is an updated function but in the current version, when you double click the bio, the bio content is hidden

leeannachen (commented):
i changed some files with my last merge (mostly for html/css) but you might want to run it things again after pulling to make sure things are still routed properly

proj/src/main/webapp/db.js
leeannachen: nvm i just saw you commented on the issue

proj/src/main/webapp/script.js
```diff
@@ -365,327 +31,6 @@ function appendMessage(payload){
...
-    else{
-        console.log("Error. Geolocation not supported or not enabled");
-    }
-    return;
-}
-
-
 /*
     Chatroom sidebar
  */
```
aokeke, line 36: Should this section also be part of chat.js?

aokeke (approved):
LGTM, but one question.

proj/src/main/webapp/script.js
arringtonh: we're trying to separate the files based on the SDK (auth + db/location/notifications which will get a script at some point) and since the chatroom sidebar stuff doesn't need auth + db we're leaving it here

PR #47, opened 2020-07-13
Adds functions to update and add user tags

nlumiere:
Additionally adds field of 'keyIdDict' to user which must be referenced in order to remove users from chats. This prevents other users without access to admin chat information to remove users from a chat.

### DIFF
--- a/proj/src/main/webapp/script.js
+++ b/proj/src/main/webapp/script.js
@@ -91,6 +91,18 @@ function findChatAndAddUser(snapshot){
     }
 }
 
+function setUserTags(tags){
+
+}
+
+function addUserTags(tags){
+
+}
+
+function removeUserFromChatByTag(tag){
+
+}
+
 // Create or join chatroom
 function createOrJoinChat(currentTag){
     var ref = firebase.database().ref("/chat/");
@@ -132,7 +144,7 @@ if(btnSignUp){
             for(var ii = 0; ii < tagList.length; ii++){
 
                 var tag = tagList[ii];
-                var key = await createOrJoinChat(tag)
+                var key = await createOrJoinChat(tag);
                 allTags[tag] = key;
             }
 
@@ -274,7 +286,6 @@ function initUserChat(){
             resolve(1);
         });
     });
-
 
 }
 // Sets title of page
--- a/proj/src/main/webapp/script.js
+++ b/proj/src/main/webapp/script.js
@@ -171,8 +171,10 @@ if(btnLogout){
 firebase.auth().onAuthStateChanged(firebaseUser => {
     if(firebaseUser){
         console.log(firebaseUser);
-        userTagsRef = firebase.database().ref("users/" + firebaseUser.uid + "/alltags");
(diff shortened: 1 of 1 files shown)


proj/src/main/webapp/script.js
```diff
@@ -60,27 +64,26 @@ function createNewChatWithUser(tag){
     var currentReference = firebase.database().ref("/chat/" + tag);
     var postKey = currentReference.push(newChat).key;
     currentReference = firebase.database().ref("/chat/" + tag + "/" + postKey + "/users/");
-    addUserToTag(currentReference);
+    addUserToTag(currentReference, tag);
 
     return postKey;
 }
 
 MAX_CHAT_SIZE = 200;
```
aokeke: This can be removed since it is at the top now.

proj/src/main/webapp/script.js
```diff
@@ -27,12 +29,14 @@ if(btnLogin){
     });
 }
 
+var keyIdDict = {};
```
aokeke, line 33: I don't like this global `keyIdDict`. Is it possible to do this functionality without global variables like this. Like if this were an input to methods or was returned by methods.

proj/src/main/webapp/script.js
```diff
@@ -27,12 +29,14 @@ if(btnLogin){
...
 
 // Adds user to an existing chat when given a reference to the place in the database
-function addUserToTag(reference){
+function addUserToTag(reference, tag){
     var currentUID = firebase.auth().currentUser.uid;
```
aokeke: `const currentUID=...` instead of `var`?

proj/src/main/webapp/script.js
```diff
@@ -27,12 +29,14 @@ if(btnLogin){
...
     console.log("adding new user to chat room with uid: " + currentUID);
-    reference.push(currentUID);
+    var removalKey = reference.push(currentUID).key;
+    keyIdDict[tag] = removalKey;
```
aokeke, line 40: Rather than storing it to a dict here and then somewhere else adding the key to the user, can you add it to the user here so that all the functionality of adding a user to a chat is in one method? Although it might require multiple calls to the database rather than one call at the end of everything.

aokeke: Scratch that, I guess that is what `createOrJoinChat` does. But I think this would be better if it returned the key instead of saving it to a global object. Maybe similar to how it is done for `allTags`.

proj/src/main/webapp/script.js
```diff
@@ -183,6 +192,148 @@ firebase.auth().onAuthStateChanged(firebaseUser => {
     }
 });
 
+
+/* Set by authState listener    */
+var allTagsRef;
+var abridgedTagsRef;
+var tagRemovalRef;
+var abridgedTagRemovalRef;
```
aokeke: same comment here about global variables. I think having everything defined within the scopes they are used is safer so that if different parts of the page are loading at different times, there is no chance of them modifying the same reference.

aokeke (changes requested):

aokeke (approved):
LGTM

PR #48, opened 2020-07-13
allows messages to be submitted with enter key

leeannachen:
debugs the issue with sending message with enter key, the function was not being called. putting the messaging function first works better because this function is a primary function of the chat.  thus even if following functions have a bug, loading and sending messages should not be a problem

### DIFF
--- a/proj/src/main/webapp/script.js
+++ b/proj/src/main/webapp/script.js
@@ -233,6 +233,9 @@ function init() {
     
     auth.onAuthStateChanged(async firebaseUser => {
         if(firebaseUser){
+
+            clickWithEnterKey();
+
             // InitUserChat sets information relevant to logged-in user
             // Must run before enclosed functions
             await initUserChat().then(function(){
@@ -241,8 +244,6 @@ function init() {
                 populateProfileSidebar(firebaseUser);
                 initBio();
             });
-
-            clickWithEnterKey();
         }
         else{
             window.location.replace("/static/welcome.html");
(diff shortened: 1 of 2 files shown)


proj/src/main/webapp/static/chat.html
```diff
@@ -173,7 +173,7 @@ <h1 class="display-4 card-header" id="big-title"> Chat Room </h1>
                         <div class="input-group-append">
                             <span class="input-group-text attach_btn"><i class="fas fa-paperclip"></i></span>
                         </div>
-                        <textarea id="message-input" class="form-control type_msg" placeholder="Type your message..."></textarea>
+                        <textarea id="message-input" class="form-control type_msg" id="message-input" placeholder="Type your message..."></textarea>
```
aokeke: This isn't needed because it already has the id at the start of the tag.

proj/src/main/webapp/script.js
```diff
@@ -233,6 +233,9 @@ function init() {
     
     auth.onAuthStateChanged(async firebaseUser => {
         if(firebaseUser){
+
+            clickWithEnterKey();
+
```
aokeke, line 238: Now that there is more on the page than just the chat, you might need to revisit how this listener is done. If they are typing their bio and press enter, shouldn't it submit the bio not the chat? Maybe find which thing is in focus in `clickWithEnterKey` before submitting the chat?

aokeke (changes requested):

proj/src/main/webapp/static/chat.html
leeannachen: oop didn't catch that, deleting the duplicate right now

proj/src/main/webapp/script.js
leeannachen: the function listens for the enter key within the text box; if your are pressing enter anywhere else on the page, the message will not be sent

aokeke (approved):
LGTM

PR #50, opened 2020-07-13
Allow users to change bio on double click.

arringtonh:


leeannachen (commented):
lgtm; i am assuming that the static bio will get override by the one from the database

aokeke (commented):
I don't think I see how this does anything with double clicking. Can you elaborate on how this works?

arringtonh:
there were already existing functions that allowed users to change their bio, it looks like the text field for the bio got deleted though so all I had to do was add it back with the same ID @aokeke

aokeke (approved):
LGTM

PR #51, opened 2020-07-13
Fixes various bugs addressed in project tab

nlumiere:


proj/src/main/webapp/script.js
```diff
@@ -429,10 +433,10 @@ function initUserChat(){
 
 }
 // Sets title of page
-function setTitle(){
-    nameRef = firebase.database().ref("/chat/"+tag+"/"+globalChatId+"/chatInfo/name");
+function setTitle(dbRefObj){
+    nameRef = dbRefObj.parent.child('chatInfo');
```
nlumiere: Before anyone comments on this, the dbRefObj was the wrong child of where I needed, but this lets me reuse that object

proj/src/main/webapp/script.js
```diff
@@ -623,7 +631,7 @@ function populateSidebar() {
 
 function changeChatOnClick(domElement, tag, chatId) {
     domElement.addEventListener('click', function() {
-        var dbRefObject = getDbRef(tag, chatId);
+        var dbRefObject = getDbRef(newTag, chatId);
```
arringtonh: this line would revert the duplication bug, it should be tag and not newTag

arringtonh (changes requested):

proj/src/main/webapp/script.js
aokeke: Where is newTag defined? Should this be tag?

proj/src/main/webapp/script.js
```diff
@@ -139,6 +139,10 @@ if(btnSignUp){
             var allTags = {};
             for(var ii = 0; ii < tagList.length; ii++){
 
+                while(tagList[ii].charAt(ii) == ' '){
+                    tagList[ii] = tagList[ii].substr(1);
+                }
+
```
aokeke: This looks like it is trying to remove leading spaces or maybe trailing spaces? Using `charAt(ii)` looks like a mistake, though because `ii` is the index of the tag we are looking at.

Also this should be done a couple lines above this where `tagList[ii].trim()` is called.

aokeke (changes requested):

aokeke (approved):
LGTM

arringtonh (approved):

proj/src/main/webapp/script.js
```diff
@@ -217,8 +217,14 @@ async function setUserTags(tagList){
...
-        const allTagsRef = firebase.database().ref(abridgedTagsRef);
-        const tagRemovalRef = firebase.database().ref(abridgedTagRemovalRef);
+        
+        var allTagsRef;
+        var tagRemovalRef;
+        await new Promise(function(resolve){
+            allTagsRef = firebase.database().ref(abridgedTagsRef);
+            tagRemovalRef = firebase.database().ref(abridgedTagRemovalRef);
+            resolve(1);
+        });
```
aokeke, line 227: Could this also be achieved with
```
const allTagsRef = await firebase.database().ref(abridgedTagsRef);	
const tagRemovalRef = await firebase.database().ref(abridgedTagRemovalRef);
```

PR #53, opened 2020-07-14
shows the most recent messages first and a profile pic placeholder

leeannachen:


proj/src/main/webapp/script.js
```diff
@@ -64,3 +64,7 @@ function SidebarCollapse () {
     // Collapse/Expand icon
     $('#collapse-icon').toggleClass('fa-angle-double-left fa-angle-double-right');
 }
+
+    // scroll down chat history to show recent message
+    var chatHistory = document.getElementById("message-list");
+    chatHistory.scrollTop = chatHistory.scrollHeight;
```
aokeke: Why is this over here by itself outside of any function?

aokeke (changes requested):

showermat (approved):

proj/src/main/webapp/script.js
leeannachen: just deleted it; it was the change that got accidentally left behind when merging with the new chat.js

aokeke (approved):
LGTM

PR #54, opened 2020-07-14
Fix duplicate preview bug.

arringtonh:


nlumiere (commented):
LGTM, happy to have this resolved

leeannachen (commented):
lgtm

aokeke (approved):
LGTM. 
I think in the future it would be good to move toward a session cache for storing information that needs to be global like which chat is currently active rather than having a global variable floating around. That  way all of these global variables will be collected together and visible in one place.

PR #55, opened 2020-07-14
Seperates auth/database and rest into two different files

nlumiere:


arringtonh (commented):
lgtm

aokeke (changes requested):
I see a large amount of code being removed from scripts.js and I see chat.js has some stuff, but I don't see where the auth code went. Is there a file missing from this PR?

nlumiere:
> I see a large amount of code being removed from scripts.js and I see chat.js has some stuff, but I don't see where the auth code went. Is there a file missing from this PR?

The auth code is at the beginning, I only copied over the parts of scripts.js which are more recent than chat.js. Check the code of branch 'bugfixing' and it will be there. @aokeke

aokeke (approved):
LGTM

PR #56, opened 2020-07-14
prevent blank message inputs

leeannachen:


proj/src/main/webapp/chat.js
```diff
@@ -428,23 +428,29 @@ function initRef(dbRefObject) {
 
 function pushChatMessage() {
     const messageInput = document.getElementById('message-input');
-
-    var message = {
+    // prevent blank messages
+    if (messageInput.value.trim().length != 0){
+        var message = {
         content : messageInput.value,
```
aokeke, line 433: I'm not sure if it is what you want, but maybe you should include the `.trim()` here also when you're storing it. It means you can't have messages prefixed or suffixed with white space which can either be good or bad. Just for you to think about.

proj/src/main/webapp/chat.js
```diff
@@ -428,23 +428,29 @@ function initRef(dbRefObject) {
...
         timestamp : new Date().getTime(),
         senderDisplay : firebase.auth().currentUser.displayName,
         senderUID : firebase.auth().currentUser.uid
```
aokeke, line 436: nit: can you indent this block.

aokeke (approved):
This looks like it has some changes that are might also be present in your other PR. After you merge that one, pull master into this branch to clean up the extra diffs. You might see some merge conflicts.

proj/src/main/webapp/chat.js
leeannachen: this maybe very niche but i think trimming spaces would mess up those ASCII pic/art people occasionally send in chat

proj/src/main/webapp/chat.js
leeannachen: for sure; i indented this in the branch i am currently working on

PR #59, opened 2020-07-15
add mobile and tablet view for sign up page

leeannachen:
tablet view (iPad): 
-viewing port is for devices smaller than an iPad Pro and larger than a cell phone
Screenshot 2020-07-15 at 3 05 36 PM

mobile view (iPhone X):
-hides the welcome graphics and shows only the sign in form 
image


arringtonh (commented):
lgtm

proj/src/main/webapp/static/signup.html
```diff
@@ -1,57 +1,60 @@
...
+            <h4>meet friends • engage in communities</h4>
           </div>
           <div class="center-info">
             <div id="chat-image-containter">
-              <img id="chat-image"src="./images/friendsChatting.svg" alt="Friends chatting icon">
+              <img id="chat-image"src="./images/friendsChatting.svg" alt="Friends chatting icon" class="b-margin-3">
             </div>
             <div class="inblock-center-container">
-              <h3 class="black-bold-subtitle center-text"> meet friends <br> create communities</h3>
+              <h3 class="center-text" id="info-message"> spend less time finding chat rooms <br> and more time making friends</h3>
```
aokeke: I kind of like the other slogan a bit more because it was short and simple.

proj/src/main/webapp/static/signup.html
```diff
@@ -61,18 +64,24 @@ <h4>welcome</h4>
               <input type="text" id="email" name="email" class="input" required><br>
               <label for="password">password:</label><br>
               <input type="password" id="pass" name="password" class="input" required><br>
+              <label for="password">confirm password:</label><br>
+              <input type="password" id="passconfirm" name="password" class="input" required><br>
```
aokeke: Is there any logic for making sure this actually matches the password?

aokeke (changes requested):

proj/src/main/webapp/static/signup.html
aokeke: You don't actually have to change this though.

proj/src/main/webapp/static/signup.html
leeannachen: working on this on another pull request but there isn't any logic here

proj/src/main/webapp/static/signup.html
leeannachen: interesting, we can definitely talk about it in the team meeting tomorrow. i was thinking what it meant for a person that doesn't know what the app does but short and simple slogan are more impactful

aokeke (approved):
LGTM

PR #61, opened 2020-07-16
ensures that the password and confirmed password match 

leeannachen:


aokeke (approved):
LGTM, we should also have some sort of logic to make sure the passwords are long enough. And it might be good to give some text explaining why the passwords are underlined in red. Like "Passwords entered don't match." or something.

leeannachen:
> LGTM, we should also have some sort of logic to make sure the passwords are long enough. And it might be good to give some text explaining why the passwords are underlined in red. Like "Passwords entered don't match." or something.

sounds good i will keep working it to set password requirements and popup for errors

nlumiere (approved):
lgtm

showermat (approved):

PR #62, opened 2020-07-16
set up password restrictions 

leeannachen:
the password must at least be 8 character and contains at least one number and symbol 
below the password box is a visual checklist that turns each requirement box green upon fulfillment 
Screenshot 2020-07-16 at 6 16 31 PM

### DIFF
--- a/proj/src/main/webapp/script.js
+++ b/proj/src/main/webapp/script.js
@@ -77,10 +77,8 @@ function meetRequirements(){
     strong password and matching confirmation password **/
     var goodPassword = [passwordLength,containsNumber,containsSymbol].every(function(handler){return handler(password)})
     if (goodPassword){
-        console.log("good password");
         return true;
     } else{
-        console.log("bad password");
         return false;
     }
 }
@@ -91,7 +89,6 @@ function passwordLength(password){
         return true;
     } else{
         $('#pwLength').removeClass('alert-success');
-        console.log("too short");
         return false;
     }
 }
@@ -103,7 +100,6 @@ function containsNumber(password){
         return true;
     } else{
         $('#pwNumber').removeClass('alert-success');
-        console.log("no number");
         return false;
     }
 }
@@ -121,6 +117,7 @@ function containsSymbol(password){
     }
 }
 
+
 /* Check for password confirmation 
     enable button if and only if the password meets the requiremetns and match **/
 $('#pass, #passconf').on('keyup', function(){
@@ -129,12 +126,14 @@ $('#pass, #passconf').on('keyup', function(){
             $('#btnSignUp').prop('disabled', false);
             $('#pass').css('border-bottom','2px solid #d1b280');
             $('#passconf').css('border-bottom','2px solid #d1b280');
+            noMatch.hidden = true;
         } else{
             // disable sign up button
             $('#btnSignUp').prop('disabled', true);
             // underline the inputs in red
-            $('#pass').css('border-bottom','3px solid #fa8072');
-            $('#passconf').css('border-bottom','3px solid #fa8072');
+            $('#pass').css('border-bottom','2px solid #fa8072');
+            $('#passconf').css('border-bottom','2px solid #fa8072');
+            noMatch.hidden = false;
         }
     }
 })
--- a/proj/src/main/webapp/script.js
+++ b/proj/src/main/webapp/script.js
@@ -72,10 +72,6 @@ $('#pass').on('keyup', function(){
 function meetRequirements(){
     const password = document.getElementById("pass");
     var goodPassword = false;
-    passwordLength(password);
-    containsNumber(password);
-    containsSymbol(password);
-
     /* returns true or false so that the sign up button is enables with a 
     strong password and matching confirmation password **/
     var goodPassword = [passwordLength,containsNumber,containsSymbol].every(function(handler){return handler(password)})
--- a/proj/src/main/webapp/script.js
+++ b/proj/src/main/webapp/script.js
@@ -60,44 +60,55 @@ function getLocation() {
 /*
     Setting password 
 */ 
-$('#pass').click(function(){
-    // activates the password requirments 
-    showRequirements();
-});
-
-function showRequirements(){
-    console.log("show requirements");
-}
 
 // Check if the password meets the password requirement 
 $('#pass').on('keyup', function(){
     const password = document.getElementById("pass");
-    containsNumber();
-    containsSymbol();
-});
-
-// function passwordLength(){
-//     if 
-// }
-
-function containsNumber(){
-    var number = /[0-9]/;
-    if(password.value.match(number)){
-        $('#pwNumber').addClass('alert-success');
+    var goodPassword = false;
+    passwordLength(password, goodPassword);
+    containsNumber(password, goodPassword);
+    containsSymbol(password, goodPassword);
+    // enables sign up button if the pw requirements are met
+    if (goodPassword === true) {
+        $('#btnSignUp').prop('disabled', false);
     } 
     else{
-        $('#pwNumber').removeClass('alert-success');
+        $('#btnSignUp').prop('disabled', true);
+    }
+});
+
+function passwordLength(password){
+    if (password.value.length > 7){
+        $('#pwLength').addClass('alert-success');
+        goodPassword = true; 
+    } 
+    else{
+        $('#pwLength').removeClass('alert-success');
+        goodPassword = false;
     }
 }
 
-function containsSymbol(){
+function containsNumber(password){
+    var number = /[0-9]/;
+    if(password.value.match(number)){
+        $('#pwNumber').addClass('alert-success');
+        goodPassword = true;
+    } 
+    else{
+        $('#pwNumber').removeClass('alert-success');
+        goodPassword = false;
+    }
+}
+
+function containsSymbol(password){
     var symbol = /[$-/:-?{-~!"^_`\[\]]/;
     // const password = document.getElementById("pass");
     if(password.value.match(symbol)){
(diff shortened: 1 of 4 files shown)


proj/src/main/webapp/script.js
```diff
@@ -50,17 +50,85 @@ function errorCallback(err){
...
+    Setting password 
+*/ 
+
+// Check if the password meets the password requirement 
+$('#pass').on('keyup', function(){
+    const password = document.getElementById("pass");
+    var goodPassword = false;
+    passwordLength(password, goodPassword);
+    containsNumber(password, goodPassword);
+    containsSymbol(password, goodPassword);
```
aokeke: Did you mean to return goodPassword from these three functions and do something like `goodPassword &=passwordLength(password)`

proj/src/main/webapp/script.js
```diff
@@ -50,17 +50,85 @@ function errorCallback(err){
...
+    // enables sign up button if the pw requirements are met
+    if (goodPassword === true) {
```
aokeke: I think you can just do `if (goodPassword){...}` here and use JS truthy logic. Checking for `===` equality isn't usually what you want.

proj/src/main/webapp/script.js
```diff
@@ -50,17 +50,85 @@ function errorCallback(err){
...
+    }
+});
+
+function passwordLength(password){
+    if (password.value.length > 7){
+        $('#pwLength').addClass('alert-success');
+        goodPassword = true; 
+    } else{
+        $('#pwLength').removeClass('alert-success');
+        goodPassword = false;
```
aokeke: goodPassword isn't defined in this method. It looks like it is supposed to be an input, but as I mentioned above, it might be better to return this value instead. Passing in a boolean for mutation seems a bit risky.

proj/src/main/webapp/script.js
aokeke: Alternatively you can do something like
`var goodPassword = [passwordLength,containsNumber,containsSymbol].every(function(handler){return handler(password)})` if you want to get fancy.

proj/src/main/webapp/script.js
```diff
@@ -50,17 +50,85 @@ function errorCallback(err){
...
+// Check for password confirmation
+$('#pass, #passconf').on('keyup', function(){
+    if ((pass.value.length != 0) && (passconf.value.length != 0)){
+        if ($('#pass').val() == $('#passconf').val()){
+            $('#btnSignUp').prop('disabled', false);
+            $('#pass').css('border-bottom','2px solid #d1b280');
+            $('#passconf').css('border-bottom','2px solid #d1b280');
+        } else{
+            // disable sign up button
+            $('#btnSignUp').prop('disabled', true);
```
aokeke, line 125: I think the sign up button should also be disabled if the passconf button's length is 0. Here if you have two matching passwords then erase all of passconf at once, it might jump to not doing this check and keep the button enabled.

proj/src/main/webapp/static/signup.html
```diff
@@ -81,7 +94,7 @@ <h5 class="text-center"><a href="./welcome.html">already have an account? sign i
   <footer class="hide-on-mobile">
     <div>icons made by <a href="https://www.flaticon.com/authors/freepik" title="Freepik">freepik</a> from <a href="https://www.flaticon.com/" title="Flaticon">www.flaticon.com</a></div>
   </footer>
-  <script src="../chat.js"></script>
   <script src="../script.js"></script>
+  <script src="../chat.js"></script>
```
aokeke, line 102: was this necessary?

aokeke (changes requested):

proj/src/main/webapp/script.js
leeannachen: there is a required html tag on both inputs so the form cannot be submitted if the passconf's length is 0

proj/src/main/webapp/static/signup.html
leeannachen: yup chat.js have all of the datastore files and functions

proj/src/main/webapp/script.js
leeannachen: also firebase seems to throw their own error if the password is too short

proj/src/main/webapp/script.js
leeannachen: gotcha, revising right now

proj/src/main/webapp/script.js
leeannachen: ah okay, changed

proj/src/main/webapp/script.js
leeannachen: haha definitely going to implement the fancy one

proj/src/main/webapp/script.js
leeannachen: also the function that enables the sign up button also verifies that both fields are not empty

proj/src/main/webapp/script.js
```diff
@@ -50,17 +50,99 @@ function errorCallback(err){
...
+});
+
+// Check if the password is strong enough 
+function meetRequirements(){
+    const password = document.getElementById("pass");
+    var goodPassword = false;
+    passwordLength(password);
+    containsNumber(password);
+    containsSymbol(password);
+
```
aokeke: Can you remove this chunk since it isn't necessary anymore.

proj/src/main/webapp/script.js
```diff
@@ -50,17 +50,99 @@ function errorCallback(err){
...
+    /* returns true or false so that the sign up button is enables with a 
+    strong password and matching confirmation password **/
+    var goodPassword = [passwordLength,containsNumber,containsSymbol].every(function(handler){return handler(password)})
+    if (goodPassword){
+        console.log("good password");
+        return true;
+    } else{
+        console.log("bad password");
+        return false;
+    }
```
aokeke: If you don't mind losing the `consol.log(...)`, you can just `return goodPassword;`

proj/src/main/webapp/script.js
aokeke: Ok, SGTM, if there are any password combinations that firebase doesn't like, or if firebase throws an error when you try to create an account, it would be good to display some sort of message to the user rather than silently failing in the background and not moving to the next screen. But that is something for a later changelist and it applies to all communications with the database. Maybe creating a generic on error handler that will create a pop up saying something like "an issue occurred".

proj/src/main/webapp/static/signup.html
aokeke: I mean was the reordering of the imports necessary. 

But now that you mention it, i don't see why chat.js is being used by the sign up page. If you were trying to make each page's .js file independent, I'd either recreate the bits needed for the sign up page in the sign up page's .js file or create a base .js file that both can inherit functionality from. But that might be a task for a different PR.

aokeke (changes requested):

proj/src/main/webapp/script.js
leeannachen: isn't this necessary to call those function? (the functions also change visual parts on the page depending on the password input)

proj/src/main/webapp/script.js
leeannachen: just deleted :)

proj/src/main/webapp/script.js
leeannachen: makes sense, especially for errors like invalid password or email already in use

leeannachen:
mobile view:
*all error/info boxes are hidden when the page first loads 
image

proj/src/main/webapp/script.js
aokeke: It shouldn't be necessary. In the `var goodPassword = [passwordLength,containsNumber,containsSymbol].every(function(handler){return handler(password)})` each of them are getting called, so you don't need to call them again up here.

proj/src/main/webapp/script.js
```diff
@@ -50,17 +50,100 @@ function errorCallback(err){
...
+    containsSymbol(password);
+
+    /* returns true or false so that the sign up button is enables with a 
+    strong password and matching confirmation password **/
+    var goodPassword = [passwordLength,containsNumber,containsSymbol].every(function(handler){return handler(password)})
+    if (goodPassword){
+        return true;
+    } else{
+        return false;
+    }
```
aokeke: I think you can just `return goodPassword;` and save yourself 4 lines of code.

aokeke (changes requested):

proj/src/main/webapp/script.js
leeannachen: gotcha, it's changed

proj/src/main/webapp/script.js
```diff
@@ -50,17 +50,96 @@ function errorCallback(err){
...
+    meetRequirements();
+    pwLength.hidden = false;
+    pwNumber.hidden = false;
+    pwSymbol.hidden = false;
+});
+
+// Check if the password is strong enough 
+function meetRequirements(){
+    const password = document.getElementById("pass");
+    var goodPassword = false;
```
aokeke: this `var goodPassword=false;` also can be removed.

proj/src/main/webapp/script.js
aokeke: I don't think i see the change.

aokeke (changes requested):

proj/src/main/webapp/script.js
leeannachen: deleted

proj/src/main/webapp/script.js
leeannachen: okay i pushed my repo again

aokeke (approved):
LGTM

PR #64, opened 2020-07-18
getting the config

aokeke:


nlumiere (changes requested):
Just keeping this here so I don't accidentally merge this until I've gotten all useful info out of it

PR #65, opened 2020-07-20
Let users block other users.

arringtonh:
hides messages from blocked users. currently no way to unblock users but we'll get there

### DIFF
--- a/proj/src/main/webapp/chat.js
+++ b/proj/src/main/webapp/chat.js
@@ -633,6 +633,39 @@ function friendRequestButton(uid) {
     });
 }
 
+function blockButton(uid) {
+    const currUserRef = firebase.database().ref('/users/'+currentUID+'/blocked/'+uid);
+    const button = document.getElementById('block');
+
+    currUserRef.off();
+    currUserRef.on('value', function(snap) {
+        if (snap.val()) {
+            // user is blocked
+            button.innerText = 'unblock user';
+            button.onclick = function() {
+                currUserRef.remove();
+                button.onclick = null;
+            }
+        } else {
+            // user not blocked
+            button.onclick = function() {
+                currUserRef.set(true);
+                // remove from each others' friend lists
+                const currFriendRef = firebase.database().ref('/users/'+currentUID+'/friends/'+uid);
+                const otherFriendRef = firebase.database().ref('/users/'+uid+'/friends/'+currentUID);
+
+                currFriendRef.once('value', function(snap) {
+                    if (snap.val()) {
+                        currFriendRef.remove();
+                        otherFriendRef.remove();
+                    }
+                })
+                button.onclick = null;
+            }
+        }
+    })
+}
+
 function getDbRef(tag, chatId) {
     const path = "/chat/"+tag+"/"+chatId+"/messages";
     const dbRefObj = firebase.database().ref(path);
--- a/proj/src/main/webapp/chat.js
+++ b/proj/src/main/webapp/chat.js
@@ -444,11 +444,15 @@ function initRef(dbRefObject) {
     // note that when a comment is added it will display more than the limit, which
     // is intentional
     dbRefObject.off('child_added');
-    dbRefObject.orderByChild("timestamp").limitToLast(LIMIT + 1).on('child_added', snap => {
-        messageUid = snap.val().senderUID;
-        blockedRef = firebase.database().ref('/users/'+currentUID+'/blocked/'+messageUid);
-        blockedRef.once('value', function(snap2) {
-            if (!snap2.val()) { //user is not blocked
+    blockedRef = firebase.database().ref('/users/'+currentUID+'/blocked')
+    blockedRef.once('value').then(function(snap) {
+        return snap.val(); // dict of blocked users
+    }).then(function(blockedUsers) {
+        blockedUsers = blockedUsers || {};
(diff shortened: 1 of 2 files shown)


proj/src/main/webapp/chat.js
```diff
@@ -634,6 +639,39 @@ function friendRequestButton(uid) {
...
+                currUserRef.remove();
+            }
+        } else {
+            // user not blocked
+            button.innerText = 'block user';
+            button.onclick = function() {
+                currUserRef.set(true);
+                // remove from each others' friend lists
+                const currFriendRef = firebase.database().ref('/users/'+currentUID+'/friends/'+uid);
+                const otherFriendRef = firebase.database().ref('/users/'+uid+'/friends/'+currentUID);
```
aokeke: Maybe you don't want to do this? It might give people a hint that they've been blocked which I think blocked functionalities don't like to do. Then again it could just be that the user unfriended you. I'll defer to your judgement and your other reviewers as long as it is considered.

aokeke (approved):
One small thing to consider. Other than that LGTM.

proj/src/main/webapp/chat.js
```diff
@@ -634,6 +639,39 @@ function friendRequestButton(uid) {
     });
 }
 
+function blockButton(uid) {
+    const currUserRef = firebase.database().ref('/users/'+currentUID+'/blocked/'+uid);
+    const button = document.getElementById('block');
+
+    currUserRef.off();
+    currUserRef.on('value', function(snap) {
```
aokeke: Looking at this again, it looks like the watch on this database reference hangs around even after you change close this window.  It is closed and reopened when you open this window. If you're looking at a lot of users profiles, you will get a build up of watchers which could slow the browser down. Is there a way to call currUserRef.off(); when the profile panel closes?

aokeke (changes requested):

proj/src/main/webapp/chat.js
arringtonh: fixed

proj/src/main/webapp/chat.js
```diff
@@ -434,14 +432,24 @@ function initRef(dbRefObject) {
...
     dbRefObject.off('child_added');
-    dbRefObject.orderByChild("timestamp").limitToLast(LIMIT + 1).on('child_added', snap => {
-        if (!firstChildKey) {
-            firstChildKey = snap.key;
-        } else {
-            messageDom = createMessageWithTemplate(snap.key, snap.val());
-            chat.appendChild(messageDom);
-        }
-    });
+    blockedRef = firebase.database().ref('/users/'+currentUID+'/blocked')
```
aokeke: I'm not sure if this line will be resolved by the other PR, or if it will cause merge conflicts, but can this `currentUiD` be found inside the function instead of pulled from the global variable.

proj/src/main/webapp/chat.js
```diff
@@ -469,36 +477,37 @@ function pushChatMessage() {
...
-        if (messages.hasOwnProperty(key)) {
-            if (!firstChildKey) {
-                firstChildKey = key;
-            } else {
-                const messageDom = createMessageWithTemplate(key, messages[key]);
-                chat.insertBefore(messageDom, firstChild);
-            }
-        }
+    if (messages.scrollTop === 0) {
+        const blockedRef = firebase.database().ref('/users/'+currentUID+'/blocked');
```
aokeke: same thing here with currentUID

proj/src/main/webapp/chat.js
```diff
@@ -615,6 +631,38 @@ function friendRequestButton(uid) {
     });
 }
 
+function blockButton(uid) {
+    const currUserRef = firebase.database().ref('/users/'+currentUID+'/blocked/'+uid);
```
aokeke: Same here with currentUID

proj/src/main/webapp/chat.js
```diff
@@ -615,6 +631,38 @@ function friendRequestButton(uid) {
...
+                currUserRef.remove();
+            }
+        } else {
+            // user not blocked
+            button.innerText = 'block user';
+            button.onclick = function() {
+                currUserRef.set(true);
+                // remove from each others' friend lists
+                const currFriendRef = firebase.database().ref('/users/'+currentUID+'/friends/'+uid);
+                const otherFriendRef = firebase.database().ref('/users/'+uid+'/friends/'+currentUID);
```
aokeke: and here and here.

proj/src/main/webapp/chat.js
```diff
@@ -729,8 +777,11 @@ function addUserInfoToDom(userObj) {
     const profile = document.getElementById('user-profile');
     if (userObj.uid !== currentUID) {
```
aokeke: This one too if possible, but I think maybe we can save it for a later PR since it isn't new.

aokeke (changes requested):
Some currrentUIDs that need to be defined so that we aren't adding more technical debt.

proj/src/main/webapp/chat.js
arringtonh: fixed for real this time. i did it once but i think a accidentally deleted the changed when switching branches -- oops

aokeke (commented):
LGTM

aokeke (approved):
LGTM!

PR #66, opened 2020-07-20
Let users add and remove tags.

arringtonh:
Chat duplication glitch seems to get worse with this one

### DIFF
--- a/proj/src/main/webapp/chat.js
+++ b/proj/src/main/webapp/chat.js
@@ -706,14 +706,13 @@ function initBio() {
 
 function populateSidebar() {
     const userTagsRef = firebase.database().ref('/users/'+currentUID+'/allTags');
-    userTagsRef.orderByKey().on('child_added', snap => {
-
-        const chatTag = snap.key;
-        const chatId = snap.val();
-
-        // get last message
-        makePreviewWithLastMessage(chatTag, chatId)
-
+    userTagsRef.orderByKey().on('value', snap => {
+        const sidebar = document.getElementById('chats-submenu');
+        sidebar.innerHTML = '';
+        for (tag in snap.val()) {
+            chatId = snap.val()[tag];
+            makePreviewWithLastMessage(tag, chatId)
+        }
     });
 }
 
--- a/proj/src/main/webapp/chat.js
+++ b/proj/src/main/webapp/chat.js
@@ -757,7 +757,7 @@ function addUserInfoToDom(userObj) {
 
     for (tag in userObj.tags) {
         if (userObj.tags.hasOwnProperty(tag)) {
-            addTag(tag);
+            addTag(tag, userObj.uid);
         }
     }
 }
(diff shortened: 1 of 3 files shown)


proj/src/main/webapp/chat.js
```diff
@@ -283,8 +282,8 @@ async function setUserTags(tagList){
 async function addUserTags(tagList){
     if(firebase.auth().currentUser){
 
-        const abridgedTagsRef = "/users/" + firebaseUser.uid + "/allTags";
-        const abridgedTagRemovalRef = "/users/" + firebaseUser.uid + "/tagRemovalDict";
+        const abridgedTagsRef = "/users/" + currentUID + "/allTags";
+        const abridgedTagRemovalRef = "/users/" + currentUID + "/tagRemovalDict";
```
aokeke: I'm not sure where `currentUID` is defined. Also having it defined globally, isn't a great idea. Can we replace it wth `firebase.auth().currentUser.uid` everywhere it is used so that functions are not dependent on the global variable wherever it is set.

aokeke (changes requested):

proj/src/main/webapp/chat.js
arringtonh: fixed

proj/src/main/webapp/chat.js
```diff
@@ -381,13 +380,12 @@ function init() {
...
     const chat = document.getElementById('chatbox');
     chat.addEventListener('scroll', addMoreMessagesAtTheTop);
-
 }
 
 
 function initUserChat(){
-    currentUID = firebase.auth().currentUser.uid;
-    const userTagsRef = firebase.database().ref('/users/'+currentUID+'/allTags');
+    currentUid = firebase.auth().currentUser.uid;
```
aokeke: `const` missing in front. Without it, it might try to assign some global variable of currentUid which may not exist instead of the local one.

proj/src/main/webapp/chat.js
```diff
@@ -734,13 +738,72 @@ function addUserInfoToDom(userObj) {
...
+// adds tag input
+function addTagsToDom(uid) {
+    tagsRef = firebase.database().ref('/users/'+uid+'/allTags');
+    tagContainer = document.querySelector('.tag-container');
+
+    const tagInput = document.createElement('input');
+    tagInput.id = "tag-input";
+    tagContainer.append(tagInput);
+    tagInput.addEventListener('keyup', function(event) {
+        if (event.keyCode === 13) {
```
aokeke: can you add a comoment saying what key is 13? I think it is enter but I'm not sure.

aokeke (changes requested):
2 more tiny things.

proj/src/main/webapp/chat.js
arringtonh: fixed

proj/src/main/webapp/chat.js
arringtonh: fixed

aokeke (commented):
LGTM

aokeke (approved):
LGTM

PR #76, opened 2020-07-23
Fix build issues

aokeke:
Move the location of the package.json so that node_modules are not included every time we try to do a build, run or deploy.
Update the instructions in the README.md given the file changes.
Create a short script for deploying locally and deploying to the project.


leeannachen (approved):
lgtm

PR #77, opened 2020-07-24
create mobile and tablet view for the chatroom page

leeannachen:
desktop view:
image

 tablet view:
image

mobile view:
image
image


proj/src/main/webapp/script.js
```diff
@@ -144,24 +144,64 @@ $('#collapse-icon').addClass('fa-angle-double-left');
...
         SeparatorTitle.addClass('d-flex');
     }
-    
-    // Collapse/Expand icon
-    $('#collapse-icon').toggleClass('fa-angle-double-left fa-angle-double-right');
-}
\ No newline at end of file
+
+    // move the buttom on mobile view
+    if ((screen.width<500) && ($( "#sidebar-container" ).hasClass( "sidebar-collapsed" ))) {
```
aokeke: 500 seems like a bit of a magic number for a .js file. Is that the size of a phone or something? It might be good to have this in a constant somewhere so it can be referenced wherever it is needed. But not necessary.

proj/src/main/webapp/script.js
```diff
@@ -144,24 +144,64 @@ $('#collapse-icon').addClass('fa-angle-double-left');
...
+    }
+}
+
+$( document ).ready(function() {
+    if (screen.width < 750) {
+        checkLoadingDisplays();
+    }
+
+    // adjust message tempate proportions
+    if (screen.width < 800) {
```
aokeke, line 174: These 2 numbers as well

aokeke (approved):

proj/src/main/webapp/script.js
leeannachen: The Pixel phone has a width of 411px so I rounded it to 500px for the mobile view

proj/src/main/webapp/script.js
leeannachen: 800 is for a general tablet view while 750 for tablets smaller than an iPad

arringtonh (commented):
lgtm!

PR #78, opened 2020-07-24
Clean up chat functions

aokeke:
Add some test cases for the chat functionality and make the tagRemovalRefs no longer global.


proj/src/main/webapp/test/chat.test.js
```diff
@@ -0,0 +1,158 @@
...
+            await firebase.auth().signOut()
+        }
+    });
+
+    after(async function() {
+        if (testconfig.shouldCleanUp) {
+            // // Delete the test users
+            // for (let i=0;i<testUsers.length;i++){
+            //     await deleteUser(testUsers[i])
+            // }
```
aokeke: Considering if we should just never delete some test users so they can be reused. If you delete and recreate users too many times, firebase starts blocking your IP.

nlumiere (changes requested):
Pretty sure there are gonna be some conflicts with current master

proj/src/main/webapp/static/chat.html
```diff
@@ -157,7 +158,7 @@ <h1 class="display-4 card-header" id="big-title"> Chat Room </h1>
                                     </div>
                                     <div calss="col-11">
                                         <li class="list-group-item">
-                                            <div class=" justify-content-between">
+                                            <div class="message-header justify-content-between">
```
leeannachen: added this line to my version and it fixed all the bugs 👍

leeannachen (commented):
lgtm

proj/src/main/webapp/chat.js
```diff
@@ -236,24 +231,27 @@ async function setUserTags(tagList){
...
                 }
                 else{
-                    key = currentTags[tag];
+                    allTags[tag] = currentTags[tag];
+                    tagRemovalDict[tag] = currentTagRemovalDict[tag];
                     delete currentTags[tag];
+                    delete currentTagRemovalDict[tag];
                 }
-                allTags[tag] = key;
             }
```
nlumiere, line 181: This looks like it should work, but it's pretty convoluted still. I'm good with it as long as it's tested and confirmed to work.

nlumiere (approved):
I don't see why anything shouldn't work... Though lease test the more complicated things if you haven't already.

arringtonh (approved):
lgtm!

PR #84, opened 2020-07-25
Breakout sign up functions

aokeke:
Move the sign up functions into their own file
Fix the validation so it happens whenever the password changes
Add test cases for the sign up functions.
Fix an issue where the chat previews were not showing
Fix a minor issue where the last chat preview did not have an upper border


leeannachen (approved):
lgtm

nlumiere (approved):
LGTM. You were missing two lines when creating a user which added location data to the user in the database, but other than that looks good. (Edit: I committed the two missing lines to your branch before reviewing)
Just make sure everything is tested before merging please!

leeannachen (approved):
lgtm (edit: didn't realize I already commented)

PR #86, opened 2020-07-27
fix password requirements bug

leeannachen:
not all password functions were being called so the password error messages were not consistent with the user's input


proj/src/main/webapp/script.js
```diff
@@ -24,25 +24,37 @@ require('bootstrap'); // for collapse
...
 function meetRequirements(){
     const password = document.getElementById("pass");
-    /* returns true or false so that the sign up button is enables with a 
-    strong password and matching confirmation password **/
+    // runs each function to see if the alert css needs to be changed
+    containsNumber();
+    passwordLength();
+    containsSymbol();
+    // returns false after receiving the first false return 
     var goodPassword = [passwordLength,containsNumber,containsSymbol].every(function(handler){return handler(password)})
```
aokeke: We shouldn't need to run each one before setting goodPassword because the `.every` will call each one for us. The problem is that `.every` shortcircuits when one is false. So it won't check. if it contains a number or symbol if it the password isn't long enough. As I broke out the sign up function into its own file in https://github.com/googleinterns/step14-2020/pull/84 I also addressed this issue.

proj/src/main/webapp/script.js
```diff
@@ -69,7 +81,7 @@ function containsNumber(password){
 }
 
 function containsSymbol(password){
-    var symbol = /[$-/:-?{-~!"^_`\[\]]/;
+    var symbol = /[$-/:-?.{@-~!"^_`\[\]]/;
```
aokeke: I'll add this to https://github.com/googleinterns/step14-2020/pull/84.

proj/src/main/webapp/welcome.js
```diff
@@ -31,7 +31,7 @@ function initLoginButtons(){
             const promise = auth.signInWithEmailAndPassword(emailVal, passVal).then(function(){
                 window.location.replace("chat.html");
             });
-            promise.catch(e => console.log(e.message));
+            promise.catch(e => alert(e.message));
```
aokeke, line 34: I'll also add this to https://github.com/googleinterns/step14-2020/pull/84

proj/src/main/webapp/script.js
leeannachen: sorry i just updated the regex again to include ASCII characters 58-64 and 34-47

leeannachen: `var symbol = /[!-/:-@{-~^_\[\]]/;`

arringtonh (commented):
lgtm

proj/src/main/webapp/script.js
leeannachen: gotcha, i'll look at your PR

leeannachen:
bug fixed with merge #84

PR #92, opened 2020-07-29
Fixing some bugs

aokeke:
Fix firebase.storage not defined error by requiring firebase/storage.
Fix chat longitude not updating after a user leaves the chat.
Fix chat dom being update with messages from old chat after changing to a new chat.


leeannachen (approved):
lgtm

arringtonh (approved):
lgtm

PR #93, opened 2020-07-29
enhance sign up functionality 

leeannachen:
-displaying signup errors as pop ups
-adjusted password requirements
-set a requirement for at least one tag


proj/src/main/webapp/signup.js
```diff
@@ -8,30 +8,50 @@ function initSignUp(){
...
+//             // Initialize auth object
+//             const auth = firebase.auth();
+
+//             const promise = auth.signInWithEmailAndPassword(emailVal, passVal).then(function(){
+//                 window.location.replace("chat.html");
+//             });
+//             promise.catch(e => alert(e.message));
+//         });
+//     }
+
```
aokeke: Can this be removed?

proj/src/main/webapp/signup.js
```diff
@@ -8,30 +8,50 @@ function initSignUp(){
...
+                    lat = 999;
+                    long = 999;
+                }
+                await signUp(fname.value, lname.value, txtEmail.value, txtPassword.value, tagStr.value, lat, long);
             }
-            else{
-                lat = 999;
-                long = 999;
+            catch (e) {
+                alert(e.message);
```
aokeke: Does this need to be wrapped in a try/catch? What happens if it isn't? Can the error catching happen closer to where the error actually gets thrown.

proj/src/main/webapp/signup.js
```diff
@@ -96,7 +116,7 @@ function meetRequirements(password){
 
 function passwordLength(password){
     let alertObj = document.getElementById("pwLength");
-    return styleAlert(alertObj,password.length > 7);
+    return styleAlert(alertObj,password.length > 5);
```
aokeke, line 104: Can you also update the failing test case.

proj/src/main/webapp/signup.js
```diff
@@ -127,10 +147,11 @@ function initializePasswordValidation(){
     const noMatch = document.getElementById("noMatch");
     const btnSignUp = document.getElementById("btnSignUp");
     const pass = document.getElementById("pass");
-    const passconf = document.getElementById("passconf");
+    const passconf = document.getElementById("passconf");
```
aokeke: Can you remove the extra whitespace here?

proj/src/main/webapp/signup.js
```diff
@@ -143,15 +164,35 @@ function initializePasswordValidation(){
...
+        // this is to prevent the page from being visually overwhelming and help guide the user
+        if (noMatch.hidden){
+            $('#tags').css('border','3px solid #fa8072');
+        }
+        return false;
+    }
+}
+
+function enableSignUp(passVal){
+    $('#tags').keyup(hasTag);
```
aokeke: Since hasTag is being called on the next line, and enableSignUp is already called in the keyUp function for #tags,#pass,#passconf, we don't need this line. If this is here, then everytime we press a key, we'll add a new listener and hastags will get called an extra time unnecessarily.

proj/src/main/webapp/signup.js
```diff
@@ -143,15 +164,35 @@ function initializePasswordValidation(){
...
 
+function hasTag(){
+    const tagBox = document.getElementById("tags");
+    if (tagBox.value.trim().length != 0){
+        $('#tags').css('border','none');
+        return true;
+    } else {
+        // the tag box will only be in red after the user has input a valid password 
+        // this is to prevent the page from being visually overwhelming and help guide the user
+        if (noMatch.hidden){
```
aokeke, line 165: I don't think noMatch is defined in a scope where this function can see it. You might need to re-find noMatch in the DOM if you want to use it here.

proj/src/main/webapp/signup.js
```diff
@@ -143,15 +164,35 @@ function initializePasswordValidation(){
...
+            $('#tags').css('border','3px solid #fa8072');
+        }
+        return false;
+    }
+}
+
+function enableSignUp(passVal){
+    $('#tags').keyup(hasTag);
+    if (meetRequirements(passVal) && noMatch.hidden && hasTag()){
```
aokeke: If you remove the line above as suggested in the above comment, hasTags will only get called if `noMatch.hidden` is true AND `meetsRequirements(passVal)` is true. So if hasTags has added the popup and then you go back and make the password invalid, then adding tags will not get rid of the pop up.

The reason is because `&&` shortcircuits  as soon as one of the values is false. If you use `&` then all three will still be called no matter if the first ones are true or false. If you use & instead of && make sure to leave a comment explaining why because it is uncommon to see and someone might come along later and think it is ok to add the extra &.

proj/src/main/webapp/static/signup.html
```diff
@@ -57,7 +57,7 @@ <h4 class="m-text" id="welcome">welcome</h4>
               <!-- password tests -->
                 <div class="alert alert-secondary" role="alert" id="pwLength" hidden>
                     <span class="glyphicon glyphicon-exclamation-sign" aria-hidden="true"></span>
-                     At least 8 characters—the more characters, the better 
+                     At least 6 characters—the more the merrier
```
aokeke, line 60: the comma disappeared

proj/src/main/webapp/style.css
```diff
@@ -134,8 +134,9 @@ h5 a:visited {
 .sign-in {
   justify-content: center;
   padding-left: 5%;
-  padding-right: 5%;
+  padding-right: 5%;
```
aokeke: Can you remove this whitespace?

aokeke (changes requested):

proj/src/main/webapp/signup.js
leeannachen: I think so; I wasn't sure if the original author was intending to come back to this

aokeke: since it is the loginBtn, I think this is in the welcome.js file. If it was already in this file, I think I might have done that by mistake. Github shows it as an addition, though.

aokeke (changes requested):

proj/src/main/webapp/signup.js
```diff
@@ -127,10 +147,11 @@ function initializePasswordValidation(){
...
     const btnSignUp = document.getElementById("btnSignUp");
     const pass = document.getElementById("pass");
-    const passconf = document.getElementById("passconf");
+    const passconf = document.getElementById("passconf"); 
     /* Check for password confirmation 
-        enable button if and only if the password meets the requiremetns and match **/
-    $('#pass, #passconf').on('keyup', function(){
+        enable button if and only if the password meets the 
+        requirements and match each other**/
+    $('#pass, #passconf, #tags').on('keyup', function(){
```
showermat, line 139: Wait, why are we checking the password on tag changes?

proj/src/main/webapp/static/signup.html
showermat: It's fine.

proj/src/main/webapp/signup.js
showermat: This smells like debugging code to me.  Is the error message user-oriented?  (That is, could the user take action based on this message?)  If not, it would be better to catch it in the most specific scope possible, and make some graceful UI change rather than just throwing up an alert.

showermat (changes requested):

proj/src/main/webapp/signup.js
leeannachen: gotcha i'll remove this

proj/src/main/webapp/signup.js
leeannachen: the alert is target for the user so they can see what information is missing or incorrect. and i think i understand what you guys mean, i am wrapping the try catch around the signup function

proj/src/main/webapp/signup.js
leeannachen: yup

proj/src/main/webapp/signup.js
leeannachen: ooh, thank you. i didn't know about `&`. i will change that right now

proj/src/main/webapp/style.css
leeannachen: yup

proj/src/main/webapp/signup.js
leeannachen: all three things are checked because the enableSignUp function checks if all three satisfies the requirements before enabling the sign up button. enableSignUp uses the validity from checking if the passwords match and the presence of tags

proj/src/main/webapp/signup.js
leeannachen: will do

proj/src/main/webapp/signup.js
leeannachen: you are right, it seems like it is not defined in the `hasTag` function but it does work.  for security, i will declare it within the scope

proj/src/main/webapp/signup.js
```diff
@@ -171,7 +171,7 @@ function hasTag(){
 
 function enableSignUp(passVal){
     $('#tags').keyup(hasTag);
-    if (meetRequirements(passVal) && noMatch.hidden && hasTag()){
+    if (meetRequirements(passVal) && noMatch.hidden & hasTag()){
```
aokeke: Can you also add a comment about why we are using single & instead of &&?  Also can you either add parentheses around the first two or switch the first && to & also so that weird order of operations stuff doesn't happen where hasTags won't get called.

proj/src/main/webapp/signup.js
```diff
@@ -13,6 +13,7 @@ function initSignUpButtons(){
     // Add sign up event
     if(btnSignUp){
         btnSignUp.addEventListener("click", async function () {
+
```
aokeke: Can we remove this extra space just to minimize the diff

proj/src/main/webapp/test/signup.test.js
```diff
@@ -99,7 +99,7 @@ describe('Test password validations', function() {
         assert.isTrue(signup.meetRequirements('testing1!'))
     });
     it('password too short', async function() {
-        assert.isFalse(signup.meetRequirements('test1!'))
+        assert.isFalse(signup.meetRequirements('test'))
```
aokeke: Can you add back the number and the symbol and remove some letters instead? The goal of this is to test the password being too short, and if we remove the symbols, we might see this check returning false because the password is missing a symbol or a number rather than because the password is too short.

proj/src/main/webapp/signup.js
aokeke: bump this comment.

proj/src/main/webapp/signup.js
```diff
@@ -31,7 +32,12 @@ function initSignUpButtons(){
                 lat = 999;
                 long = 999;
             }
-            await signUp(fname.value, lname.value, txtEmail.value, txtPassword.value, tagStr.value, lat, long);
+            try {
+                await signUp(fname.value, lname.value, txtEmail.value, txtPassword.value, tagStr.value, lat, long);
+            }
+            catch (e) {
+                alert(e.message);
+            }
```
aokeke, line 39: Since there is already a catch within signUp (about lines 81-84ish), is this still needed?

aokeke (changes requested):

proj/src/main/webapp/test/signup.test.js
leeannachen: oh gotcha didn't think of that

proj/src/main/webapp/signup.js
leeannachen: sure thing

proj/src/main/webapp/signup.js
leeannachen: sorry, i missed this comment before, i have changed it

proj/src/main/webapp/signup.js
leeannachen: That alert doesn't show weak passwords and other registration issues

proj/src/main/webapp/signup.js
leeannachen: the `enableSignUp` function checks for matching password requirements along with the presence of tags. if any of the fields change, the signup button will disabled or enables accordingly

showermat: Both the comment above and the name of the function on line 107 are misleading, then.  This function clearly does more than check whether the password contains a number!  I think they should be updated to reflect that.

showermat (changes requested):

proj/src/main/webapp/signup.js
aokeke: Oh I see, in that case, can you instead put on line 84ish a catch instead of this here? The. way this is now, if there is an issue with splitting tags or something, this will hide it. I want the catch to be constrained to the actual user creation (lines 54-84)
something like `await auth.createUserWithEmailAndPassword(email, pass).then(async function(){... 
complicated logic for setting up chats ...}).catch(function(err){alert(err);});`

aokeke (changes requested):

proj/src/main/webapp/signup.js
aokeke: I think you're confused Matt. This method is called `initializePasswordValidation`, not `containsnumber`

showermat (approved):
Never mind me.

aokeke (approved):
LGTM

PR #96, opened 2020-07-29
Adds message handling; adds topic subscription and message sending courtesy of @aokeke

nlumiere:
Notifications work and they look nice


proj/src/main/webapp/notifications.js
```diff
@@ -105,12 +105,14 @@ function sendNotificationForChat(message) {
...
-    notificationBody = name+' has sent a message:\n'+message.content
+    notificationBody = name+' has sent a message:\n\n'+message.content
     let payload = {
         "to": getTopic(tag,chatId),
         "notification": {
             "title": tag,
-            "body": notificationBody
+            "body": notificationBody,
+            "icon": '/static/images/friendsChatting.png',
+            "click_action": 'https://8080-9f73b32e-6ccb-426e-8d6e-c03d6f324666.us-west1.cloudshell.dev/static/chat.html'
```
aokeke, line 115: This url will open just the chat rather than opening a specific chat. Unfortunately we don't have something built in to redirect them to a specific chat yet, though , so this is fine.

proj/src/main/webapp/notifications.js
```diff
@@ -105,12 +105,14 @@ function sendNotificationForChat(message) {
...
     const name = message.senderDisplay;
-    notificationBody = name+' has sent a message:\n'+message.content
+    notificationBody = name+' has sent a message:\n\n'+message.content
     let payload = {
         "to": getTopic(tag,chatId),
         "notification": {
             "title": tag,
-            "body": notificationBody
+            "body": notificationBody,
+            "icon": '/static/images/friendsChatting.png',
```
aokeke, line 114: Is there a reason you use the png instead of the svg? I think the svg is better because it will resize well.

aokeke (changes requested):
One tiny thing.

aokeke (approved):
LGTM

arringtonh (approved):
lgtm

PR #99, opened 2020-07-30
fixes annoying issue where users continue to receive messages after signing out

nlumiere:


proj/src/main/webapp/chat.js
```diff
@@ -686,7 +686,9 @@ async function addUsernameToMessage(uid, preview) {
     const userRef = firebase.database().ref('/users/'+uid);
     await userRef.once("value", snap => {
         if(snap.val()) {
-            preview.querySelector('#username').innerText = snap.val().firstName + ' ' + snap.val().lastName;
+            if(uid){
+                preview.querySelector('#username').innerText = snap.val().firstName + ' ' + snap.val().lastName;
+            }
             // add pfp
             const url = snap.val().photo || DEFAULT_PFP;
             if (sessionStorage[uid+" pfp"]) {
```
showermat, line 694: Since `uid` is used here, are we sure we don't want it to be in the if statement as well?

PR #104, opened 2020-07-31
Add ability to chat one-on-one with friends

arringtonh:


nlumiere (commented):
lgtm for now, before too long please add a call to notifications.subscribeToTagChatId so users can receive notifications from 1 : 1 chats 👍

proj/src/main/webapp/chat.js
```diff
@@ -1001,6 +1052,15 @@ function addFriendToDom(uid) {
             })
         }
 
+        // sub to notifications
+        notifications.subscribeToTagChatId(tag, chatId);
```
aokeke: I think this needs to be a couple lines lower. `tag` is not defined until a couple lines later.

proj/src/main/webapp/test/chat.test.js
```diff
@@ -14,6 +14,77 @@ require("../app.js");
...
+        password: 'test123!'
+    }];
+    let chatId;
+    const oneOnOneRef = firebase.database().ref('/chat/chats-1on1');
+
+    before(async function() {
+        for (let i = 0; i < testUsers.length; i++) {
+            await testHelper.createUserIfNotExisting(testUsers[i]);
+            await firebase.auth().signInWithEmailAndPassword(testUsers[i].email, testUsers[i].password);
+            testUsers[i]["uid"] = firebase.auth().currentUser.uid;
```
aokeke: Check out `testHelper.createUserIfNotExisting`. It updates the testUser objects to include the uid after creating the user and it leaves the user signed in when it finishes. So you will probably need to sign out afterward so that the next user can properly sign in and have it's uid logged. I guess you have `if (i==1) {...} else { sign out }` below

proj/src/main/webapp/test/chat.test.js
```diff
@@ -14,6 +14,77 @@ require("../app.js");
...
+            if (i == 1) {
+                chatId = oneOnOneRef.push().key;
+                await chat.createFriendChat(chatId, testUsers[0]["uid"]);
```
aokeke: Can you move this part into an `it("...",function() {...})" test case and do some checks that the database properly created the 1:1 chat properly?

proj/src/main/webapp/test/chat.test.js
```diff
@@ -14,6 +14,77 @@ require("../app.js");
...
+    });
+
+    after(async function() {
+        if (testconfig.shouldCleanUp) {
+            // Remove nodes created
+            await testHelper.emptyDatabase();
+        }
+
+        // delete created chat
+        oneOnOneRef.child(chatId).remove();
```
aokeke: This shouldn't be needed with `testHelper.emptyDatabase();`

proj/src/main/webapp/test/chat.test.js
```diff
@@ -14,6 +14,77 @@ require("../app.js");
...
+
+        // sign out
+        await firebase.auth().signOut();
```
aokeke: I'm wondering if this should be part of the test case, but I don't have any strong feelings because it is good to make sure you are signed out when everything is finished. It might have an error if somehow the test case already signs you out though, but the same thing is true if it is in the test case and you somehow get signed out before you explicitely tell it to sign you out.

proj/src/main/webapp/test/chat.test.js
```diff
@@ -14,6 +14,77 @@ require("../app.js");
...
+        })
+
+        it("second user's friend value is set to the chatId", async function() {
+            const uid1 = testUsers[0].uid;
+            const uid2 = testUsers[1].uid;
+            const friendRef = firebase.database().ref('/users/'+uid2+'/friends/'+uid1);
+            await friendRef.once('value', function(snap) {
+                assert.equal(snap.val(), chatId);
+            })
+        })
```
aokeke, line 86: If it isn't too challenging, can you try adding a test case that writes to the 1:1 chat and checks that the message ends up in the correct place in the database?

aokeke (changes requested):

proj/src/main/webapp/test/chat.test.js
arringtonh: yeah I have to leave one user logged in because the function i'm testing uses firebase.auth().currentUser.uid

proj/src/main/webapp/test/chat.test.js
arringtonh: i don't think i'll be able to do this unless i don't use pushChatMessage and do it manually, since pushChatMessage interacts with the DOM

proj/src/main/webapp/chat.js
```diff
@@ -421,7 +421,7 @@ function pushChatMessage() {
         }
         // push message to datastore
         let dbRefObj = getActiveDbRef();
-        const messageKey = dbRefObj.push(message).key;
+        dbRefObj.push(message).key;
```
aokeke: do you need the `.key` at the end still?

proj/src/main/webapp/chat.js
```diff
@@ -557,8 +557,12 @@ function friendRequestButton(uid) {
 
                     const currFriendRef = firebase.database().ref('/users/'+currentUid+'/friends/'+uid);
                     const otherFriendRef = firebase.database().ref('/users/'+uid+'/friends/'+currentUid);
-                    currFriendRef.set(true);
-                    otherFriendRef.set(true);
+
+                    const oneOnOneRef = firebase.database().ref('/chat/chats-1on1');
```
aokeke: `chats-1on1` should probably be a constant in case it needs to be referenced in multiple places. Like how it is used again in `addFriendToDom`

aokeke (changes requested):
A couple more small comments.

proj/src/main/webapp/chat.js
arringtonh: nope. fixed

leeannachen (commented):
lgtm

aokeke (approved):
LGTM

PR #105, opened 2020-07-31
Sends user verification emails to account when creating account not on dev server

nlumiere:


proj/src/main/webapp/chat.js
```diff
@@ -318,18 +318,33 @@ const LIMIT = 20; // how many messages to load at a time
...
-                initRef();
-                populateProfileSidebar(firebaseUser.uid);
-                addBlockedToSettings();
-            });
+            // If the user is verified or using a test server
+            if(firebaseUser.emailVerified || window.location.href.includes("https://8080")){
+                if(firebaseUser.emailVerified)
+                    console.log("You have been verified as a Camaraderie user");
+                else
+                    console.log("You have been verified as a Camaraderie testing developer");
```
aokeke: Can you put curly braces around the parts in the if/else. It probably doesn't need it, but just for my sanity.

proj/src/main/webapp/chat.js
```diff
@@ -318,18 +318,33 @@ const LIMIT = 20; // how many messages to load at a time
...
-            // InitUserChat sets information relevant to logged-in user
-            // Must run before enclosed functions
-            await initUserChat().then(function(){
-                populateSidebar();
-                initRef();
-                populateProfileSidebar(firebaseUser.uid);
-                addBlockedToSettings();
-            });
+            // If the user is verified or using a test server
+            if(firebaseUser.emailVerified || window.location.href.includes("https://8080")){
```
aokeke, line 322: It seems weird to have something for the devs in the deployed code. Is there any other way to do this? Like maybe we can check if the user has a specific exemption property in the database that the devs can manually add from the firebase database page?

aokeke (approved):
LGTM but a couple comments.

leeannachen (approved):
lgtm

proj/src/main/webapp/chat.js
nlumiere: Ideally we would be able to create admin accounts and whatnot, but given the very constrained time at this point I don't think it makes sense to build an admin/dev account feature.

PR #109, opened 2020-08-03
Fix profile picture uploading; hide profile upload on other profiles.

arringtonh:


proj/src/main/webapp/chat.css
```diff
@@ -1,5 +1,12 @@
 @import url('https://fonts.googleapis.com/icon?family=Material+Icons');
 
+/* styling for hiding something */
+.hidden {
+    position: absolute;
+    left: -999px;
+    bottom: -999px;
```
aokeke: why not use `display:hidden` or `display:none`?

proj/src/main/webapp/chat.js
```diff
@@ -722,23 +722,31 @@ async function addUsernameToMessage(uid, preview) {
...
+    await pfpStorageRef.put(pfp).then(function() {
+        uploadStatus.innerText = 'upload success';
+        setTimeout(function(){ uploadStatus.innerText = ''; }, 5000);
+    }).catch(function(error) {
+        uploadStatus.innerText = 'upload failed: '+error.message;
+        setTimeout(function(){ uploadStatus.innerText = ''; }, 5000);
+    });
 
-    pfpStorageRef.getDownloadURL().then(function(url) {
+    await pfpStorageRef.getDownloadURL().then(function(url) {
```
aokeke: I think if you really wanted to, you could chain this with the previous await. It might remove the need for async and. awaits.
Something like 
```
pfpStorageRef.put(pfp).then(function() {
      ...
    }).catch(function(error) {
      ...
    }).then(function(){
      return pfpStorageRef.getDownloadURL();
    }).then(function(url){
      ...
    })
```

proj/src/main/webapp/chat.js
```diff
@@ -887,13 +895,15 @@ function addUserInfoToDom(userObj) {
     tagContainer.innerHTML = '';
     
     if (userObj.uid !== currentUid) {
-            friendRequestButton(userObj.uid);
-            blockButton(userObj.uid);
+        friendRequestButton(userObj.uid);
+        blockButton(userObj.uid);
+        document.getElementById('change-pfp').classList.add('hidden');
```
aokeke: Just like a couple lines below, can you use the `.hidden=true;` instead of using a class.

proj/src/main/webapp/chat.js
```diff
@@ -887,13 +895,15 @@ function addUserInfoToDom(userObj) {
...
     } else {
         document.getElementById('friend-request').hidden = true;
         document.getElementById('block').hidden = true;
         initBio();
         addTagsToDom(currentUid);
+        document.getElementById('change-pfp').classList.remove('hidden');
```
aokeke: same thing here

aokeke (changes requested):

proj/src/main/webapp/chat.css
arringtonh: i tried using display hidden, display none, and .hidden = true but none of them worked for some reason

proj/src/main/webapp/chat.js
```diff
@@ -722,23 +722,31 @@ async function addUsernameToMessage(uid, preview) {
...
     const input = document.getElementById("pfp-upload");
     const pfp = input.files[0];
 
     const currentUid = firebase.auth().currentUser.uid;
     const pfpStorageRef = firebase.storage().ref(`/profile-pictures/${currentUid}/pfp.png`);
-    pfpStorageRef.put(pfp);
-
-    pfpStorageRef.getDownloadURL().then(function(url) {
+    const uploadStatus = document.getElementById('upload-status')
+    await pfpStorageRef.put(pfp).then(function() {
```
aokeke: Do you still need the await and async on this function?

arringtonh: Yes because the stuff in the chain has to execute before it does the stuff a the bottom

aokeke: Can't you just pull the stuff at the bottom into the final `.then(function(url){...})`?

aokeke (approved):
LGTM

nlumiere (approved):
LGTM

PR #112, opened 2020-08-04
Prevent overflow in chat preview from long messages

arringtonh:
I accidentally built this branch off the profile picture branch which is why it has extra commits

### DIFF
--- a/proj/src/main/webapp/chat.js
+++ b/proj/src/main/webapp/chat.js
@@ -731,8 +731,10 @@ async function pfpOnInput() {
     const uploadStatus = document.getElementById('upload-status')
     await pfpStorageRef.put(pfp).then(function() {
         uploadStatus.innerText = 'upload success';
+        setTimeout(function(){ uploadStatus.innerText = ''; }, 5000);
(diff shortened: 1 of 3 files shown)


proj/src/main/webapp/chat.js
```diff
@@ -722,23 +722,31 @@ async function addUsernameToMessage(uid, preview) {
...
         const userPfpRef = firebase.database().ref(`/users/${currentUid}/photo`);
         userPfpRef.set(pfpStorageRef.toString());
 
         userPfp = document.getElementById('user-pfp');
         userPfp.src = url;
     })
     input.files = null; // clear the input
-    sessionStorage[currentUid +" pfp"] = null; // clear session storage
+    input.value = "";
+    delete sessionStorage[currentUid +" pfp"]; // clear session storage
```
aokeke, line 749: This change looks like it is from the pfp branch. Is this dependent on those changes?

arringtonh: no, i accidentally branched off that branch. i only made small changes so i'll just make another pr

PR #114, opened 2020-08-05
Adds 'unread' markers for new messages. 

nlumiere:
This is a functional, yet (recommended to be) temporary marker. It would be cool to see this marker stylistically updated to match google chat for example.


proj/src/main/webapp/chat.js
```diff
@@ -478,7 +488,7 @@ function addMoreMessagesAtTheTop() {
                     const messageUid = child.val().senderUID;
 
                     if (!blockedUsers[messageUid] && child.key !== firstChild.id) {
-                        const message = createMessageWithTemplate(child.key, child.val());
+                        const message = createMessageWithTemplate(child.key, child.val(), currentUid, 9999999999999);
```
showermat: Hmm.  I'm not sure what Arinze thinks, but I think it would be cleaner to use a value like 0 and special-case that in the logic, rather that use a solution that will break in...hmm...November 2286.  That could also be specified as the default argument of the function, since I think the "default" case is not to mark messages unread.

proj/src/main/webapp/chat.js
```diff
@@ -526,6 +536,9 @@ function createMessageWithTemplate(key, messageObj) {
             message.querySelector('#pfp').src = src;
         })
     })
+    if(messageObj.timestamp > lastVisited && messageObj.senderUID != firebase.auth().currentUser.uid){
+        message.querySelector('.unread').innerText = " (New)";
```
showermat: This is likely to be more useful to Lee or whoever cleans it up, but I think the standard way to do this is to have a CSS class that describes how you want to display unread messages, and then just apply that class to these messages.  That way, the styling can be changed without touching the code.

proj/src/main/webapp/firebase-messaging-sw.js
```diff
@@ -22,17 +22,17 @@ firebase.initializeApp({
...
 
 messaging.setBackgroundMessageHandler(function(payload) {
-  console.log('[firebase-messaging-sw.js] Received background message ', payload);
-  const notificationTitle = 'Background Message Title';
-  const notificationOptions = {
-    body: 'Background Message body.',
-    icon: '/firebase-logo.png'
-  };
+    console.log('[firebase-messaging-sw.js] Received background message ', payload);
+    const notificationTitle = 'Background Message Title';
```
showermat: Is it not easy to finish implementing these notification values?

showermat (commented):
Nice, I think this is almost perfect.

proj/src/main/webapp/chat.js
aokeke: I agree with Matt. I had an experience where we wrote some code once thinking something along the lines of, this'll probably change by 2018 or 2019 or something and then that date came and it didn't change and then all our test cases randomly started failing and it took so much time to figure out why things were working one day and then just suddenly stopped.

proj/src/main/webapp/chat.js
```diff
@@ -452,6 +457,11 @@ function pushChatMessage() {
             'timestamp': message.timestamp
         });
         notifications.sendNotificationForChat(message)
+
+        const unreadClasses = document.querySelectorAll('.unread');
+        unreadClasses.forEach(unreadMarker => {
+            unreadMarker.innerText = '';
+        });
```
aokeke: Do you also need to update the lastVisted value in the database here?

proj/src/main/webapp/chat.js
aokeke: Maybe just a slightly different background color for unread messages as opposed to adding text?

proj/src/main/webapp/chat.js
```diff
@@ -918,7 +931,20 @@ function changeChatOnClick(domElement, tag, chatId) {
...
+        const dbRef = firebase.database().ref("/users/" + uid + "/lastVisited/");
+        dbRef.once('value', snap => {
+            lastVisited = snap.child(tag).val();
+            initRef(lastVisited);
+        }).finally(function(){
+            var date = new Date();
+            var timestamp = date.getTime();
+            var updates = {};
+            updates[tag] = timestamp;
+            dbRef.update(updates);
```
aokeke, line 965: Should this be updating the timestamp for the current tag or the tag we are leaving or neither?

proj/src/main/webapp/notifications.js
```diff
@@ -137,4 +135,5 @@ exports.initNotifications = initNotifications;
 exports.subscribeToTagChatId = subscribeToTagChatId;
 exports.unsubscribeFromTagChatId = unsubscribeFromTagChatId;
 exports.sendNotificationForChat = sendNotificationForChat;
-exports.unSubscribeFromAllChats = unSubscribeFromAllChats;
\ No newline at end of file
+exports.unSubscribeFromAllChats = unSubscribeFromAllChats;
+exports.getToken = getToken;
```
aokeke: I don't think you need to export getToken.

aokeke (changes requested):
Just a few changes.

proj/src/main/webapp/chat.js
showermat: I agree that's probably the better stylistic approach, but the beautiful thing about applying a class is that you can then do either just by changing some CSS.

proj/src/main/webapp/chat.js
nlumiere: It's currently updates the lastVisited property for the tag that's clicked on after the dom updates

proj/src/main/webapp/chat.js
nlumiere: It doesn't really make a difference. Either way, any new messages will be more recent than the last time the user has viewed the page, and they'll receive notifications when we want them to. I was thinking we might as well not update when we don't have to and save some client-side compute time, no?

showermat (approved):
Approved in anger.

aokeke (approved):
LGTM

PR #115, opened 2020-08-05
fix mobile chat room layout 

leeannachen:
-readjust the message template for devices smaller than an ipad so that the profile pictures are not cut off 
-hide collapsed sidebar on mobile view 
-create a mobile navigation button and hide it for tablet and desktop view


aokeke (approved):
LGTM just some commented out. sections that can be removed

nlumiere (approved):
Perf

PR #116, opened 2020-08-05
Do input checking on name.

arringtonh:


leeannachen (commented):
lgtm

proj/src/main/webapp/chat.js
```diff
@@ -754,7 +754,9 @@ async function addUsernameToMessage(uid, preview) {
     const userRef = firebase.database().ref('/users/'+uid);
     await userRef.once("value", snap => {
         if(snap.val()) {
-            preview.querySelector('#username').innerText = snap.val().firstName + ' ' + snap.val().lastName;
+            const fName = snap.val().firstName || ' ';
+            const lName = snap.val().lastName || ' ';
```
aokeke, line 758: Is there a reason you use a space instead of an empty string?

aokeke (approved):
LGTM

proj/src/main/webapp/chat.js
arringtonh: no

PR #120, opened 2020-08-06
allow users to change their passwords

leeannachen:


proj/src/main/webapp/chat.js
```diff
@@ -1035,12 +1035,53 @@ function addUserInfoToDom(userObj) {
...
+        errors.push("Your password must contain at least one symbol."); 
+    }
+    if (errors.length > 0) {
+        // Shows what the new password is missing, if any
+        alert(errors.join("\n"));
+        return false;
+    }
+    return true;
+}
+
```
aokeke, line 1079: Rather than recoding the logic for verifying the password, I would reuse the logic that is in signup. However I think signup imports chat so you would have a circular dependency. Is it possible to redirect them to a page similar to sign up or maybe move the password validation things to its own module separate from signup similar to how notifications and locations does things so it can be brought into both chat and signup?

aokeke (changes requested):

proj/src/main/webapp/chat.js
leeannachen: i thought about that and was considering importing just the meetRequirements function but that only return a boolean. i felt the validation function was too small for a separate module because the one in signup is more focused on triggering styling changes on each key entered while this is basic version for a quick password change. lmk what you think

aokeke: I still think it's better to reuse existing code. That way if we decide the password should be 7 letters instead of 6, we won't accidentally forget to change it in one of the places. Or if we notice the regex is missing a character in one of the places we won't forget to update both places.

aokeke (approved):
I still prefer breaking it out into a module just so that we have one source of truth for what a valid password is, but I'll LGTM this.

PR #122, opened 2020-08-07
Protects project

nlumiere:
Removes credentials, api keys and database information. All info pasted to pod notes


aokeke (changes requested):
Also in the testing directory there was a file that had some of this information for the nlumiere project.

